PR #3, opened 2018-12-10
Improve pinentry handling

robinkrahl:
This pull requests contains several patches that improve pinentry handling.  The major changes are:

- Use a more specific cache ID to allow querying different PINs (e. g. user PIN and admin PIN).
- Display error messages in the pinentry dialog (if possible) instead of the command line.

### DIFF
--- a/nitrocli/src/pinentry.rs
+++ b/nitrocli/src/pinentry.rs
@@ -21,7 +21,7 @@ use error::Error;
 use std::process;
 
 
-const CACHE_ID: &str = "nitrokey";
+const CACHE_ID: &str = "nitrocli:user";
 
 
 fn parse_pinentry_passphrase(response: Vec<u8>) -> Result<Vec<u8>, Error> {
--- a/nitrocli/src/pinentry.rs
+++ b/nitrocli/src/pinentry.rs
@@ -49,12 +49,14 @@ fn parse_pinentry_passphrase(response: Vec<u8>) -> Result<Vec<u8>, Error> {
 }
 
 
-pub fn inquire_passphrase() -> Result<Vec<u8>, Error> {
-  const PINENTRY_ERROR_MSG: &str = "+";
+pub fn inquire_passphrase(error_msg: Option<&str>) -> Result<Vec<u8>, Error> {
+  const PINENTRY_ERROR_MSG_EMPTY: &str = "+";
   const PINENTRY_PROMPT: &str = "PIN";
   const PINENTRY_DESCR: &str = "Please+enter+user+PIN";
 
-  let args = vec![CACHE_ID, PINENTRY_ERROR_MSG, PINENTRY_PROMPT, PINENTRY_DESCR].join(" ");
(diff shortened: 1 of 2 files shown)


nitrocli/src/pinentry.rs
```diff
@@ -21,7 +21,45 @@ use error::Error;
 use std::process;
 
 
-const CACHE_ID: &str = "nitrocli:user";
+/// PIN type requested from pinentry.  The available PIN types correspond to the PIN types used by
```
d-e-s-o: I believe it is good practice to make the first sentence a summary and to separate it from the body with an empty line. E.g., https://github.com/rust-lang/rust/blob/master/src/libcore/str/mod.rs#L81

nitrocli/src/pinentry.rs
```diff
@@ -21,7 +21,45 @@ use error::Error;
...
+  User,
+}
+
+
+const CACHE_ID_PREFIX: &str = "nitrocli";
+const CACHE_ID_SUFFIX_ADMIN: &str = "admin";
+const CACHE_ID_SUFFIX_USER: &str = "user";
+
+
+fn get_cache_id(pin_type: PinType) -> String {
```
d-e-s-o: I feel like those three functions should belong into an `impl` for `PinType`. Also, can you change the return type to `&'static str`? Lastly, I believe the Rust community mostly has reached consensus on not prefixing getters with `get_`. E.g.,

```rust
impl PinType {
  fn cache_id(self) -> &'static str {
    match self {
      PinType::Admin => "nitrocli:user",
      PinType::User => "nitrocli:admin",
    }
  }
  // ...
}
```

Lastly, the project uses two spaces of indentation. I'd hope we can keep things consistent. Perhaps I find the time to set up a CI for the project, I feel as if it would help now.

nitrocli/src/pinentry.rs
```diff
@@ -87,6 +87,8 @@ fn parse_pinentry_passphrase(response: Vec<u8>) -> Result<Vec<u8>, Error> {
 }
 
 
+/// Inquire a PIN of the given type from the user or return the cached passphrase, if available.
```
d-e-s-o: Same comment with respect to the comment as for the previous commit.

```rust
Inquire a PIN of the given type from the user.

This function inquires a PIN of the given type from the user or returns the cache passphrase if available.
...
```

nitrocli/src/pinentry.rs
robinkrahl: Fair points.  I agree that the code style should be consistent.  Besides the indentation width, should we stick to the [Rust style guidelines][]?

[Rust style guidelines]: https://doc.rust-lang.org/1.0.0/style/README.html

d-e-s-o: > Fair points. I agree that the code style should be consistent. Besides the indentation width, should we stick to the [Rust style guidelines](https://doc.rust-lang.org/1.0.0/style/README.html)?

Sure thing.

d-e-s-o:
Looks great, too!

PR #9, opened 2018-12-12
Port from hidapi to nitrokey

robinkrahl:
This patch set refactors the existing commands open, close and status to use the nitrokey crate instead of directly communicating with the Nitrokey device. It also removes the hid, hidapi-sys and pkg-config dependencies that are no longer required.


d-e-s-o:
Thanks for the pull request, Robin. Please allow me a couple more days to go through the commits.

nitrocli/Cargo.toml
```diff
@@ -58,4 +63,7 @@ path = "../pkg-config"
 "hidapi-sys:0.1.4" = { path = "../hidapi-sys" }
 "libc:0.2.45" = { path = "../libc" }
 "libhid:0.4.1" = { path = "../hid" }
+"libnitrokey:0.2.1" = { path = "../nitrokey" }
+"nitrokey-sys:3.4.1" = { path = "../nitrokey-sys" }
 "pkg-config:0.3.14" = { path = "../pkg-config" }
+"rand:0.4.3" = { path = "../rand" }
```
d-e-s-o: Out of curiosity, do you any plans for making the `nitrokey` crate compatible with more recent versions of `rand`? It appears they have a [comprehensive documentation](https://rust-random.github.io/book/update-0.5.html) for the upgrade, I just haven't checked how much of the `rand` code you use.

robinkrahl: Yes, I’m planning to do that soon.  It should not be that hard as `nitrokey` only uses `rand` to generate a temporary password.

nitrocli/src/main.rs
```diff
@@ -49,6 +51,19 @@ const RECV_TRY_COUNT: i8 = 40;
 const SEND_RECV_DELAY_MS: u64 = 200;
 
 
+/// Create an `error::Error` with an error message of the format `msg: err`.
+fn get_error(msg: &str, err: libnitrokey::CommandError) -> Error {
```
d-e-s-o: Clippy has the following to say:
```
97 | fn get_error(msg: &str, err: libnitrokey::CommandError) -> Error {
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&libnitrokey::CommandError`
```
I think you may want to consider making `CommandError` derive from `Copy` (there is a `missing_copy...` lint), which should resolve this problem. Meanwhile, I think we should adhere to clippy's suggestion, unless you agree with the above and have enough changes piled up already to create a new release that would be consumes here (you don't have to do that right now, obviously).

nitrocli/src/main.rs
```diff
@@ -49,6 +51,19 @@ const RECV_TRY_COUNT: i8 = 40;
...
+  let error = format!("{}: {:?}", msg, err);
+  Error::Error(error)
+}
+
+
+/// Connect to a Nitrokey Storage device and return it.
+fn get_storage_device() -> Result<libnitrokey::Storage> {
+  libnitrokey::Storage::connect().or(Err(Error::Error("Nitrokey device not found".to_string())))
```
d-e-s-o: (I understands it's a bit of a pain to identify new problems identified by clippy, but I haven't gotten around cleaning up the existing ones; still, we should avoid introducing new warnings to converge to a maintainable state eventually, even if they are just emitted by clippy and not rustc directly)

```
warning: use of `or` followed by a function call
   --> src/main.rs:105:35
    |
105 |   libnitrokey::Storage::connect().or(Err(Error::Error("Nitrokey device not found".to_string())))
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `or_else(|_| Err(Error::Error("Nitrokey device not found".to_string())))`
```

nitrocli/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 Unreleased
 ----------
+- Use the nitrokey crate for the `open` and `close` commands instead of
+  directly communication with the Nitrokey device.
```
d-e-s-o: *`communicating`

I've kept all bullet points without an explicit full stop at the end (although they may be a full sentence). Mind doing the same for the sake of consistency?

nitrocli/src/main.rs
```diff
@@ -230,64 +245,33 @@ fn status() -> Result<()> {
...
-
-      if status == nitrokey::StorageStatus::WrongPassword {
-        pinentry::clear_passphrase(PIN_TYPE)?;
-        retry -= 1;
+  let mut retry = 3;
+  let mut error_msg: Option<&str> = None;
+  loop {
+    let passphrase = pinentry::inquire_passphrase(PIN_TYPE, error_msg)?;
+    let passphrase = String::from_utf8_lossy(&passphrase);
+    match device.enable_encrypted_volume(&passphrase) {
```
d-e-s-o: Mind rebasing your changes against my top of tree `devel`? For better or worse, I've added the `unused_results` lint to the crate. It flags
```
error: unused result
   --> src/main.rs:299:5
    |
299 | /     match device.enable_encrypted_volume(&passphrase) {
300 | |       Ok(()) => return Ok(()),
301 | |       Err(err) => match err {
302 | |         libnitrokey::CommandError::WrongPassword => {
...   |
314 | |       },
315 | |     };
    | |______^
    |
```
I understand this can be a bit annoying, but it has helped me catch a bug in the past, so it may be worth it. Work around:
```
let _ = ...
```

nitrocli/CHANGELOG.md
robinkrahl: I try to; sometimes I just don’t notice things that I am so used to.

nitrocli/src/main.rs
```diff
@@ -63,185 +53,51 @@ fn get_storage_device() -> Result<libnitrokey::Storage> {
...
-      Err(err) => {
-        if retry > 0 {
-          thread::sleep(time::Duration::from_millis(SEND_RECV_DELAY_MS));
-          continue;
-        } else {
-          return Err(Error::HidError(err));
-        }
-      },
-    }
+fn get_volume_status(status: &libnitrokey::VolumeStatus) -> String {
```
d-e-s-o: Mind returning a `&'static str` to avoid an unnecessary allocation?

nitrocli/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+- Remove the `hid`, `hidapi-sys` and `pkg-config` dependencies
```
d-e-s-o: I missed this earlier. Mind using past tense in the change log?

d-e-s-o:
Hm, my comments were for the individual commits in the order I received them. Not sure if that is apparent from Github. It appears to be combining everything at the pull request level. Let me know if something is unclear.

nitrocli/src/main.rs
robinkrahl: Unfortunately I currently can’t use clippy as it is not packaged for Debian and both the installation via `rustup` and manual compilation from source fail.  I’m happy to fix any issues you mention, but currently I can’t check them myself.

Regarding the specific error: I’d rather fix this here instead of waiting for a new `nitrokey` version. I’m also still thinking about ways to improve the error handling so this code may change anyway.

d-e-s-o:
Thanks for the updated pull request and for your help in general! I've gone through everything once more and have merged the changes. Allow me to state two things that I found and fixed along the way (for later reference):
- `Port the open and close commands to libnitrokey` left tests in place that were no longer compiling. I removed them. We should keep that in mind and potentially add back similar ones once we are back on track with respect to testing (I heard back from the Nitrokey team and they will send me [at least] a Storage stick from the sounds of it; so I can add tests once I have that)
- The same commit also uses `String::from_utf8_lossy` to convert the password as it was input. I consider that dangerous. We should never modify the password the user supplied. If we can't convert we should error out. I've added a TODO.

robinkrahl:
Thanks for your review and for merging!

PR #11, opened 2018-12-14
Add a nitrocli(1) man page

robinkrahl:
The man page could also be moved to a subdirectory like `doc` or `etc` if you want to keep the main project directory clean.

### DIFF
--- a/nitrocli/doc/nitrocli.1
+++ b/nitrocli/doc/nitrocli.1
@@ -1,0 +1,27 @@
+.TH NITROCLI 1 2018-12-14
+.SH NAME
+nitrocli \- access Nitrokey devices
+.SH SYNOPSIS
+.B nitrocli
+\fBclear\fR|\fBclose\fR|\fBopen\fR|\fBstatus\fR
+.SH DESCRIPTION
+\fBnitrocli\fR provides access to Nitrokey devices.
+Currently, \fBnitrocli\fR only supports accessing the encrypted volume of a
+Nitrokey Storage.
+.SH COMMANDS
+.TP
+.B open
+Open the encrypted volume on the Nitrokey Storage.
+The user PIN that is required to open the volume is queried using 
+\fBpinentry\fR(1) and cached by \fBgpg-agent\fR(1).
+.TP
+.B close
+Close the encrypted volume on the Nitrokey Storage.
+.TP
+.B status
+Print the status of the connected Nitrokey Storage, including the SD card
+serial number, the firmware version, the encryption status, the PIN retry count
+and the status of the volumes.
+.TP
+.B clear
+Clear the passphrase cached by the \fBopen\fR command.


nitrocli/nitrocli.1
```diff
@@ -0,0 +1,30 @@
+.TH NITROCLI 1
+.SH NAME
+nitrocli \- access Nitrokey devices
+.SH SYNOPSIS
+.B nitrocli
+\fBclear\fR|\fBclose\fR|\fBopen\fR|\fBstatus\fR
```
d-e-s-o: It appears uncommon to have the subcommand appear bold. Any particular reason to do that? Or am I just looking at the wrong role model?

nitrocli/nitrocli.1
```diff
@@ -0,0 +1,30 @@
...
+.SH DESCRIPTION
```
d-e-s-o: It appears all `man` pages that I've looked into include what appears to be the date of creation and the program version. E.g., for `sed` I can see:
`TH SED "1" "March 2018" "sed 4.5" "User Commands"`

Should we include that as well?

d-e-s-o:
Yeah, I think we should put that into a `doc/` sub directory. May I ask how you created the man page? Did you just write it by hand or use a tool to convert it from something more friendly to the editor's eyes?

nitrocli/nitrocli.1
robinkrahl: `man-pages(7)` says:
> **SYNOPSIS**
> […]
> For  commands, this shows the syntax of the command and its arguments (including options);
> boldface is used for as-is text and italics are used to indicate replaceable arguments.

As the commands are “as-is text”, I chose boldface.

nitrocli/nitrocli.1
robinkrahl: We could. I left it out as I did not see a clear advantage and as it is an additional field to maintain.  If we add it, we should use the format `YYYY-MM-DD` for the date to conform to `man-pages(7)`. I’d leave out the source and manual fields as they have sensible defaults.

robinkrahl:
Previously, I’ve used POD ([example](https://git.ireas.org/nkotp/tree/nkotp.1.pod)) to generate man pages, and I saw that there is a Rust crate for the generation ([rust-clique/man](https://github.com/rust-clique/man)).

This time, I wrote the man page directly (based on [this guide](https://liw.fi/manpages/)).  In the end, it’s not that complicated once you get used to it, and it makes it easier to deal with version control.

nitrocli/nitrocli.1
robinkrahl: I added the date.

d-e-s-o:
> Previously, I’ve used POD ([example](https://git.ireas.org/nkotp/tree/nkotp.1.pod)) to generate man pages, and I saw that there is a Rust crate for the generation ([rust-clique/man](https://github.com/rust-clique/man)).
> 
> This time, I wrote the man page directly (based on [this guide](https://liw.fi/manpages/)). In the end, it’s not that complicated once you get used to it, and it makes it easier to deal with version control.

Sounds good. We can always change that at a later point in time.

d-e-s-o:
Looks great! Thanks Robin

PR #20, opened 2018-12-19
Add `argparse` and implement `otp` command

robinkrahl:
This patch series introduces the `argparse` crate for argument parsing, adds the `commands` and `options` modules and implements the `otp {set,get,clear,status}` commands.

### DIFF
--- a/nitrocli/src/main.rs
+++ b/nitrocli/src/main.rs
@@ -70,48 +70,33 @@
 
 mod commands;
 mod error;
+mod options;
 mod pinentry;
 
 use std::process;
 use std::result;
 
-use nitrokey;
-
 use crate::error::Error;
 
 type Result<T> = result::Result<T, Error>;
 
-// A macro for generating a match of the different supported commands.
-// Each supplied command is converted into a string and matched against.
-macro_rules! commands {
-  ( $str:expr, [ $( $command:expr), *] ) => {
-    match &*$str.to_string() {
-      $(
-        stringify!($command) => {
-          if let Err(err) = $command() {
-            println!("{}", err);
-            return 1
-          }
-          return 0
-        },
-      )*
-      x => {
-        println!("Invalid command: {}", x);
-        println!("Available commands: {}", stringify!( $($command)* ));
-        return 1
-      },
-    }
-  }
-}
-
 fn run() -> i32 {
-  let argv: Vec<String> = std::env::args().collect();
-  if argv.len() != 2 {
-    println!("Usage: {} <command>", argv[0]);
-    return 1;
+  let args = std::env::args().collect();
+  match options::handle_arguments(args) {
+    Ok(()) => 0,
+    Err(err) => match err {
+      Error::ArgparseError(err) => match err {
+        // argparse printed the help message
+        0 => 0,
+        // argparse printed an error message
+        _ => 1,
+      },
+      _ => {
+        println!("{}", err);
+        1
+      }
+    },
   }
-
-  commands!(&argv[1], [commands::status, commands::open, commands::close, commands::clear]);
 }
 
 fn main() {
--- a/nitrocli/src/options.rs
+++ b/nitrocli/src/options.rs
@@ -1,0 +1,151 @@
+// options.rs
+
+// *************************************************************************
+// * Copyright (C) 2018 Daniel Mueller (deso@posteo.net)                   *
+// *                                                                       *
+// * This program is free software: you can redistribute it and/or modify  *
+// * it under the terms of the GNU General Public License as published by  *
+// * the Free Software Foundation, either version 3 of the License, or     *
+// * (at your option) any later version.                                   *
+// *                                                                       *
+// * This program is distributed in the hope that it will be useful,       *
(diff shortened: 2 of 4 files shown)


d-e-s-o:
Thanks for this pull request, Robin! Unfortunately I may need a few more days until I get to looking at it in more detail.

nitrocli/src/options.rs
```diff
@@ -0,0 +1,155 @@
...
+        Command::Status => "status",
+      }
+    )
+  }
+}
+
+impl FromStr for Command {
+  type Err = ();
+
+  fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
```
d-e-s-o: It appears we are using:
1) `use std::str::FromStr` (full qualifier in `use` statement)
2) `use std::io` and
3) `std::result::Result` (full qualifier at site of use)

I'd rather we agree on one way to reference symbols. In the past I've typically used the first variant, but it seems uncommon in the community I admit that it may not be the best one. I think 2) is most common. Shall we converge on that?

nitrocli/src/commands.rs
```diff
@@ -0,0 +1,149 @@
+// commands.rs
```
d-e-s-o: It would be great to make the introduction of this file a separate commit. I've actually done that already in [wip/argparse](https://github.com/d-e-s-o/nitrocli/commits/wip/argparse) in order to make reviewing easier. Is that okay with you?

nitrocli/src/options.rs
```diff
@@ -0,0 +1,155 @@
+// options.rs
```
d-e-s-o: Would `args.rs` be a more suitable name, given that we work with `argparse`?

nitrocli/src/options.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Executes the given parser on the given arguments and handles the result.
+///
+/// This macro executed the given argument parser on the given arguments using the standard output
+/// and error.  It then handles three cases:  If the result is `Ok`, it does nothing as the
+/// arguments have been parsed successfully.  If the result is `Err(0)`, it returns `Ok(())` as the
+/// help message has been printed by `argparse`.  Otherwise it returns `Err(Error::ArgparseError)`
+/// as `argparse` failed.
+macro_rules! parse_args {
```
d-e-s-o: I guess I do understand why you have introduced a macro here, but I think that can be solved with a function and in the process be made more clear. Would it work to simply propagate the error reported by `argparse` up (i.e., extend `Error::ArgparseError` with the error code) and then make the decision as to how to exit in the `run` function?

Largely untested: [argparse.txt](https://github.com/d-e-s-o/nitrocli/files/2704772/argparse.txt)

nitrocli/src/options.rs
```diff
@@ -0,0 +1,155 @@
...
+    argparse::Store,
+    "the command to execute (clear|close|open|status)",
+  );
+  let _ = parser.refer(&mut args).add_argument(
+    "arguments",
+    argparse::List,
+    "the arguments for the command",
+  );
+  parser.stop_on_first_argument(true);
+  parser.parse_args_or_exit();
```
d-e-s-o: I'd rather not have anything exit the program in some side module. Similarly, I think we should pass in the argument vector to this function such that environment accesses are all located in the main file and dependencies are explicit.

nitrocli/src/main.rs
```diff
@@ -68,176 +68,31 @@
...
-    }
-  }
-}
-
 fn run() -> i32 {
-  let argv: Vec<String> = std::env::args().collect();
-  if argv.len() != 2 {
-    println!("Usage: {} <command>", argv[0]);
-    return 1;
+  let (command, args) = options::parse_arguments();
```
d-e-s-o: Instead of returning the command and `execute`ing it here, do you think it would be clearer to just short-circuit that, rename `parse_arguments` to `handle_arguments` and have it deal with the dispatch of the command?

robinkrahl:
Github seems to butcher e-mail replies, sorry for the mess in the last
comment.

I’ve changed the commits to address most of your points and I removed
the OTP commits.  I did not change the file name and the use directives.
If you want to change them, please let me know (or change the commit).

d-e-s-o:
> Github seems to butcher e-mail replies, sorry for the mess in the last comment.

No worries. I have to say for a platform that is hyped as much as Github is, a couple of the services it provides leave quite a few things to be desired. I find the way reviews are handled absolutely sub par. I can't select multiple lines (or perhaps I am just too incapable to use the UI properly) and reviewing individual patches is quite messy and confusing.

robinkrahl:
> No worries. I have to say for a platform that is hyped as much as Github is, a couple of the services it provides leave quite a few things to be desired. I find the way reviews are handled absolutely sub par. I can't select multiple lines (or perhaps I am just too incapable to use the UI properly) and reviewing individual patches is quite messy and confusing.

… and the issue tracker is very basic too.  It’s fine for small projects like this one, but the stuff they have to do to make it work e. g. for the Rust project is horrific (labels for all kinds of metadata, bots just to manage the issues, etc.).

PR #23, opened 2018-12-19
pinentry: Encode spaces in prompt argument

robinkrahl:
Spaces in the arguments for gpg-connect-agent’s `GET_PASSPHRASE` command have to be esaced using a plus sign.  Somehow this was missing for the prompt argument.  This patch adds escaping for the prompt so that the pinentry dialog is displayed correctly.


d-e-s-o:
Looks great, Robin. Thanks!

PR #25, opened 2018-12-23
Add otp command with clear, get, set and status subcommands

robinkrahl:
This patch series implements access to the one-time password functionality on Nitrokey devices.  It depends on pull request #20.  I want to test whether it makes sense to create pull requests that depend on other pull requests.


robinkrahl:
Thanks for merging!  I’m not too happy with that function either.  I
might reconsider the nitrokey authentication design, but I have not yet
found a better solution.

Issue #24, opened 2018-12-19
PIN reuse across multiple devices


robinkrahl:
`nitrocli` does not keep track of the device it connects to.  If a user switches between two devices without clearing the passphrase cache, `nitrocli` will try to connect to device A using the cached passphrase for device B.  This is not a major issue, but I think we could improve `nitrocli`’s behavior by
- adding the serial number (and maybe the model?) of the connected Nitrokey to the pinentry dialog, and
- making sure that we do not reuse a cached passphrase for another device.

I was thinking of two ways to implement the second point:  Firstly, we could add the serial number to the ID of the cached passphrase (e. .g `nitrocli:admin:deadbeef` instead of `nitrocli:admin`).  This would make it harder to clear the cache with the `clear` command as we either would have to keep track of the passphrase IDs or could only clear the passphrases for the current device.  Secondly, we could store the serial number together with the stored passphrase (with a separate cache ID or as a prefix for the passphrase).  But it is unclear whether `PRESET_PASSPHRASE` is applicable for custom passphrases, and it is not enabled by default.  So we cannot set the passphrase without user interaction.

d-e-s-o:
Interesting, yeah that is a bit of a problem. Thanks for pointing that out.

I am not sure I fully understand the problem you mention with the second approach:

> But it is unclear whether `PRESET_PASSPHRASE` is applicable for custom passphrases, and it is not enabled by default.

I can't follow this part. What do you mean by "applicable for custom passphrases"? That we can't store arbitrary data? And what exactly is not enabled by default?

I honestly think that the first approach is fine, where the `clear` would only affect the currently plugged-in device. Swapping devices like that is a rare case, I would argue. And even so the behavior is still reasonable.

robinkrahl:
> I can't follow this part. What do you mean by "applicable for custom passphrases"? That we can't store arbitrary data? 

The documentation for [`PRESET_PASSPHRASE`][] refers to the “cache for the specified *keygrip*”, while `GET_PASSPHRASE` and `CLEAR_PASSPHRASE` just refer to cache IDs.  I don’t know whether this is just inconsistent/outdated documentation, or whether `PRESET_PASSPHRASE` really only applies to keys managed by GPG.

> And what exactly is not enabled by default?

The `PRESET_PASSPHRASE` command:
```
$ gpg-connect-agent "PRESET_PASSPHRASE test" "/bye"
[...]

PR #29, opened 2018-12-25
Add status output for Nitrokey Pro

robinkrahl:


nitrocli/src/commands.rs
```diff
@@ -171,41 +174,30 @@ where
...
+/// Query and pretty print the common status of all Nitrokey devices.
+fn print_status(model: &'static str, device: &nitrokey::DeviceWrapper) -> Result<()> {
+  let serial_number = device
+    .get_serial_number()
+    .map_err(|err| get_error("Could not query the serial number", &err))?;
   println!(
     r#"Status:
   model:             {model}
-  smart card ID:     {id}
+  smart card ID:     0x{id}
```
d-e-s-o: Nit: Can this addition be moved into the previous commit where you factored out this function? Or am I missing something?

nitrocli/src/commands.rs
```diff
@@ -171,41 +174,30 @@ where
...
   firmware version:  {fwv0}.{fwv1}
   user retry count:  {urc}
   admin retry count: {arc}"#,
     model = model,
-    id = smartcard_id,
-    fwv0 = firmware_major,
-    fwv1 = firmware_minor,
-    urc = user_retries,
-    arc = admin_retries,
+    id = serial_number,
```
d-e-s-o: Does it make sense to differentiate between the smart card ID (for the storage device) and the serial number for the Pro? The string `smart card ID` seems to be misleading now, and I am trying to understand whether to move this part into the device specific print function or keep it here and change the string to `serial number`.

nitrocli/src/commands.rs
robinkrahl: No.  This change is needed as we replace the value from the `StorageStatus` struct (with leading `0x`) with the value from `Device::get_serial_number` (no leading `0x`) in this commit.

nitrocli/src/commands.rs
robinkrahl: I’d prefer to have this as a common field.  It is the same ID that we will show in the pinentry dialog (see #24) and that is reported by `gpg --card-status` in the same way for both models.   I’ll rename it to *serial number*.

That said, for my Nitrokey Storage, the serial number returned by `NK_get_status_storage` and `NK_device_serial_number` do not match the number reported by `gpg --card-status`.  I’ll have to investigate that.  Could you please check the result of `NK_device_serial_number` for your Storage stick?

robinkrahl:
> I believe you should also adjust the help message printed for the status command, which still talks about the storage command only.

Thanks!  I  missed that.

nitrocli/src/commands.rs
d-e-s-o: Sorry, I did not get to that today. I will get back to you tomorrow.

robinkrahl: I opened Nitrokey/nitrokey-storage-firmware#76 for this issue.

robinkrahl: No, I’ve never experienced such behavior.  But I’ve never used GnuPG on my development stick, and I only occasionally run `nitrocli` with my production stick.  And I do not use the Storage volumes at all.  So there might be interaction problems that I did not trigger with my usage pattern.  (That reminds me that we have to implement a `--debug` option that enables the debugging output from `libnitrokey`.)

d-e-s-o:
Merged. Thanks Robin!

PR #30, opened 2018-12-25
Document the otp commands

robinkrahl:
This patch adds the otp commands to the README and describes them in the nitrocli(1) man page.

----
Maybe we should have one man page per command (i. e. move the OTP stuff to nitrocli-otp(1)).

### DIFF
--- a/nitrocli/doc/nitrocli.1
+++ b/nitrocli/doc/nitrocli.1
@@ -1,28 +1,87 @@
-.TH NITROCLI 1 2018-12-26
+.TH NITROCLI 1 2018-12-27
 .SH NAME
 nitrocli \- access Nitrokey devices
 .SH SYNOPSIS
 .B nitrocli
-\fBclear\fR|\fBclose\fR|\fBopen\fR|\fBstatus\fR
+\fIcommand\fR
+[\fIarguments\fR]
 .SH DESCRIPTION
 \fBnitrocli\fR provides access to Nitrokey devices.
-Currently, \fBnitrocli\fR only supports accessing the encrypted volume of a
-Nitrokey Storage.
+It supports the Nitrokey Pro and the Nitrokey Storage.
+It can be used to access the encrypted volume and the one-time password generator.
 .SH COMMANDS
+.SS General
 .TP
-.B open
+.B nitrocli open
 Open the encrypted volume on the Nitrokey Storage.
 The user PIN that is required to open the volume is queried using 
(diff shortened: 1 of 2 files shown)


nitrocli/doc/nitrocli.1
```diff
@@ -1,27 +1,84 @@
...
+Otherwise, every two characters are interpreted as the hexadecimal value of one
+byte.
+
+\fIalgorithm\fR is the OTP algorithm to use.
+Possible values are \fBhotp\fR for the HOTP algorithm according to RFC 4226 and
+\fBtotp\fR for the TOTP algorithm according to RFC 6238 (default).
+\fIdigits\fR is the number of digits the one-time password should have.
+Allowed values are 6 and 8 (default: 6).
+\fIcounter\fR is the initial counter if the HOTP algorithm is used (default: 0).
+\fItime window\fR is the time window used with TOTP in seconds (default: 30).
```
d-e-s-o: I noticed that the fact that this duration is in seconds should probably also appear in the help text.
From `args.rs:338`
> "The time window for TOTP (default: 30)"

Mind adding that there?

nitrocli/doc/nitrocli.1
```diff
@@ -1,27 +1,84 @@
...
+List all OTP slots.
+If \fB--all\fR is not set, empty slots are ignored.
+
+.SH EXAMPLE
+.SS One-time passwords
+Configure a one-time password slot with a hexadecimal secret representation:
+    $ \fBnitrocli otp set 0 test-rfc4226 3132333435363738393031323334353637383930 --algorithm hotp\fR
+.P
+Configure a one-time password slot with an ASCII secret representation:
+    $ \fBnitrocli otp set 0 test-rfc4226 12345678901234567890 --algorithm hotp\fR
```
d-e-s-o: I think it would be beneficial to use non-hexadecimal characters in this example as well, that makes the difference between non-ASCII and `--ascii` more clear, I'd say.

robinkrahl: I chose this example because it is the test case specified in the HOTP RFC.  But I’ll add one more to illustrate the difference.

d-e-s-o: Hey @robinkrahl, do you still plan on adding that additional example?

robinkrahl: Done!

d-e-s-o:
Looks great now! Thanks!

Issue #18, opened 2018-12-19
Support the password safe


robinkrahl:
`nitrocli` should support the password safe (PWS).  The password safe is unlocked using the user password.  Once it is unlocked, any client can access the PWS without further authentication.  The PWS can be locked using the `NK_lock` function.  On the Pro, `NK_lock` does not have side effects.  On the Storage, `NK_lock` also closes the encrypted device.  I requested a lock function for the PWS without side effects (see Nitrokey/nitrokey-storage-firmware#65), but currently, there is none.  (Ideally the PWS communication would be protected by a temporary password similar to the OTP communication.)

So we have three options:
- Do not support the PWS until we can lock it without side effects.
- Do not lock the PWS after every use, leaving it unprotected until the device is removed or the machine is restarted.
- Lock the PWS after every access, potentially closing the encrypted volume.  We could try to check whether the volume is enabled and warn the user.

What do you think?

d-e-s-o:
Thanks for bringing up this issue.

> On the Storage, NK_lock also closes the encrypted device.
Ugh. Out of curiosity: do you know why that is? Is that "just" a bug or is there a rationale behind connecting the two?

> So we have three options:
> 
> * Do not support the PWS until we can lock it without side effects.
> * Do not lock the PWS after every use, leaving it unprotected until the device is removed or the machine is restarted.
> * Lock the PWS after every access, potentially closing the encrypted volume.  We could try to check whether the volume is enabled and warn the user.

I think it would help to understand whether the witnessed behavior is a bug or a feature (see above). Also, mostly for my understanding: you are doing most of the work right now and I'd like to understand your objective some more. Is your plan to expose pretty much all the functionality the device (or `libnitrokey`) supports, or are you mostly interested in implementing the features you'd have a strong use case for? If the latter, do you plan to use the PWS? (that sort of ties into your first bullet point)

I don't think I have a strong opinion about options 2 & 3, other than stating that I'd rather not have a non-obvious side-effect like this take effect. So **silently** closing the encrypted volume is something I would not do. We could ask the user to take measures to close the volume and only allow the command to go through after that. Alternatively we could also ask how to proceed or, perhaps better, avoid interactivity by adding a `--force` option that will allow the close to go through, with the side-effect of closing the encrypted volume if open.

[...]

PR #31, opened 2018-12-27
Implement the pws and lock command

robinkrahl:
As discussed in issue #18, this patch series adds support for the password safe.


d-e-s-o:
Thanks for the pull request, Robin! I'll look into it closer tomorrow.

nitrocli/src/commands.rs
```diff
@@ -450,6 +465,43 @@ pub fn otp_status(all: bool) -> Result<()> {
...
+pub fn pws_get(
+  slot: u8,
+  show_name: bool,
+  show_login: bool,
+  show_password: bool,
+  quiet: bool,
+) -> Result<()> {
+  let device = get_device()?;
+  let pws = get_password_safe(&device)?;
+  let show_all = !show_name && !show_login && !show_password;
```
d-e-s-o: I had to read the commit description twice to make sense of this logic ;)

d-e-s-o:
Aside from my previous comment (which is for the second commit, in case that is not apparent from what Github shows) the rest looks great! Thanks! Can you update the manual page along, though?

On a different note: how strongly do you feel about including this functionality into `0.2`? Given that this is a fairly isolated feature I would say we can leave it out and then merge it in later (but given that it is fairly compact I am not super opposed to including it either).

Well, actually I would propose to resurrect the `devel` branch for the purpose of having a sort of "staging" environment for things we want to eventually go into `master`. Instead of simply rebasing we would cherry pick and still have the option of amending a commit slightly or merging it with another before it lands, in case we missed something. The release would of course be cut from `master`.

I would actually want to try that out, if only to see whether a scheme like this works better. But do let me know your thoughts in this regard.

robinkrahl:
Yes, you are absolutely right.  I tried to find a comprehensive
description for the --help message but failed, so I decided to put the
explanation in the man page.  As I have a branch with man page updates
prepared on my machine, I’d ask you to merge this without the man page.
It will come soon in a separate pull request.

> On a different note: how strongly do you feel about including this
> functionality into `0.2`?

Personally, I don’t care that much as I currently do not use the PWS
myself.  On the other hand, I don’t really see a reason not to release
functionality that is available and fully implemented.

My current goal is to ship a stable* nitrocli version that supports all
basic operations with the next stable Debian release.  This means that I
have to package some nitrocli version until 2019-02-12 (soft freeze),
and the “final” version until 2019-03-12 (full freeze).  I don’t care
whether this will be version 0.2.0 or 0.12.5.

* stable as in documented, tested and working, not 1.0.0

> Well, actually I would propose to resurrect the `devel` branch for the
> purpose of having a sort of "staging" environment for things we want
> to eventually go into `master`.

That’s a good idea.  I’m using a similar approach for `nitrokey-rs` and
`nitrokey-sys-rs` (the branch is called next instead of devel).

PR #35, opened 2018-12-28
Implement `storage` subcommand

d-e-s-o:
This pull request implements a new root-level command: `storage`. Along with that it moves the existing `open` and `close` commands in the form of subcommands into that new command. Lastly, the existing root-level command `status` is split such that it now prints only status information that is common to all Nitrokey devices, while the newly introduced `storage status` subcommand takes care of printing the storage specific status.

### DIFF
--- a/nitrocli/CHANGELOG.md
+++ b/nitrocli/CHANGELOG.md
@@ -10,6 +10,8 @@ Unreleased
 - Added the `config` command for reading and writing the device configuration
 - Moved `open` and `close` commands as subcommands into newly introduced
   `storage` command
+  - Moved printing of storage related information from `status` command
+    into new `storage status` subcommand
 - Made `status` command work with Nitrokey Pro devices
 - Enabled CI pipeline comprising code style conformance checks, linting,
   and building of the project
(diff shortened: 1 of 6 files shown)


robinkrahl:
Thanks for taking care of that!  I’ll have a look at it.  –  And yes, the Github permission system for personal repositories is pretty annoying. (&lt;ad&gt;Gitlab has a much more fine-grained permissions!&lt;/ad&gt;)

robinkrahl:
I’m currently on a low-bandwidth connection and the review web interface does not load properly, so I’ll just comment directly:

The changes look very good.  I only have some minor formal suggestions:
- In the comment for the clear function, it should say *cached*, not *cache*.
- Move the `storage_*` subcommand functions below the `storage` functions for consistency.
- Add `storage status` to the command list in the readme.

d-e-s-o:
> Thanks for taking care of that! I’ll have a look at it. – And yes, the Github permission system for personal repositories is pretty annoying. (<ad>Gitlab has a much more fine-grained permissions!</ad>)

Haha, message received :-)

> The changes look very good. I only have some minor formal suggestions:
> ...

Thanks for the review! I'll address your points (I had the functions reordered previously but that made the diff look horrible; I'll introduce a dedicated commit for that)

d-e-s-o:
Merge into `devel`.

PR #36, opened 2018-12-30
Update documentation

robinkrahl:


nitrocli/doc/nitrocli.1
```diff
@@ -35,6 +35,17 @@ printed information includes the SD card serial number, the encryption
 status, and the status of the volumes.
 
 .SS One-time passwords
+The Nitrokey Pro and the Nitrokey Store support the generation of one-time
+passwords using the HOTP algorithm according to RFC 4226 or the TOTP algorithm
+according to RFC 6238.
+The required data \- a name and the secret \- is stored on slots.
```
d-e-s-o: Should that read "in slots"?

I read up a bit on hyphen usage in `*roff` and I don't think we are doing it correctly.
- `\-` is a minus sign and while that is valid for options (because a minus is the character used for them) [1] I don't think it should be used here.
- `-` is a hyphen

Now what we really want here (at least in my world view), is an em-dash, which is `\(em` [2].

In general, we probably want to change options like `--algorithm` to `\-\-algorithm`. What say you?

[1] https://liw.fi/manpages/ (ctrl-f "an unfor")
[2] http://man7.org/linux/man-pages/man7/man-pages.7.html (ctrl-f "Em-dash")

nitrocli/README.md
```diff
@@ -25,6 +25,11 @@ The following commands are currently supported:
   - set: Set an OTP slot.
   - status: List all OTP slots.
   - clear: Delete an OTP slot.
+- pws: Access the password safe (PWS).
+  - get: Query the data on a PWS slot.
```
d-e-s-o: I still think "in a PWS" slot sounds better :)

nitrocli/doc/nitrocli.1
```diff
@@ -115,6 +116,44 @@ passwords using the \fBotp get\fR command.
...
+Print the content of one PWS slot.
+\fIslot\fR is the number of the slot.
+Per default, this command prints the name, the login and the password (in that
+order).
+If one or more of the options \fB--name\fR, \fB--login\fR and \fB--password\fR
+are set, only the selected fields are printed.
+The order of the fields never changes.
+
+The fields are printed together with a label.
+Use the \fB--quiet\fR option to supress the labels and to only output the
```
d-e-s-o: *suppress

nitrocli/doc/nitrocli.1
```diff
@@ -115,6 +116,44 @@ passwords using the \fBotp get\fR command.
...
+values stored on the PWS slot.
+.TP
+\fBnitrocli pws set \fIslot name login password\fR
+Set the content of a PWS slot.
+\fIslot\fR is the number of the slot to write.
+\fIname\fR, \fIlogin\fR and \fIpassword\fR are the date to write to the slot.
```
d-e-s-o: date -> data?

robinkrahl:
Thanks for the review.  Should be fixed, just with an en-dash instead of
an em-dash (see the commit message).

d-e-s-o:
Looks great! Thanks for the changes, Robin! Merged into `devel`.

PR #37, opened 2018-12-30
Implement change-pin and unblock-pin commands

robinkrahl:


d-e-s-o:
Change set looks awesome! Thanks for implementing this functionality @robinkrahl . I'll merge it in once the one `clippy` reported "problem" is resolved.

robinkrahl:
The clippy report does not load for me.  Would you mind fixing the
issue?  Otherwise I’ll take care of it once I’m on a regular network
connection.

d-e-s-o:
Hm, looking at it again, what are your thoughts in introducing a new `pin` command @robinkrahl ? Given that we have two sub-command candidates I think that may be better than polluting the root namespace with duplication.
I can make the change if you don't have the cycles, but let me know.

d-e-s-o:
I like the idea!

robinkrahl:
I also renamed `change` to `set` for consistency.

d-e-s-o:
Latest changes look great, Robin! I've merged them in.

PR #38, opened 2018-12-30
Set the time before generating a TOTP

robinkrahl:
This pull request depends on pull request #36.


d-e-s-o:
Thanks for finishing this up :) I think you added the documentation for `--time` to the wrong sub-command in the man page. I fixed that up. I am also not sure whether we should potentially emit an error if `--time` is supplied but `hotp` is in use. I left it as is, though. If you have an opinion on that do let me know. Regardless, this change has been merged.

PR #43, opened 2019-01-01
Topic/documentation

d-e-s-o:


d-e-s-o:
@robinkrahl , would you mind having a look at the proposed changes?

nitrocli/README.md
```diff
@@ -125,6 +124,16 @@ The [Nitrokey UG][nitrokey-ug] has generously provided the necessary
 hardware for developing and testing the program.
 
 
+License
+-------
+**nitrocli** is made available under the terms of the
+[GPLv3](gplv3-tldr).
+
+See the [LICENSE] file that accompanies this distribution for the full
```
robinkrahl: I think this should be `[LICENSE](LICENSE)`.

robinkrahl:
rendering

d-e-s-o:
Thanks for having a look so quickly, Robin!

> I’m not a big fan of using the PDF – Github’s rendering is very poor, at least on my machine (I’ll attach a screenshot), it uses an image to represent text, and we have to include a generated file in the repository. But if you think that’s the best solution, I can live with it.

Yeah, I was pretty sure you'd say that :-) It looks the same on my system. I am not opposed to publishing a HTML rendered version of the man page at, say, https://linux.die.net/man/, or Debian's repo or whatever else, in addition. But I think I prefer the proposed solution as a first step because it does not rely on such a service and it is always up-to-date (well, provided the PDF is up-to-date but that is something that can be reasonably easy locally verified/enforced). I was also considering creating some automation around the tagging of releases and automatically creating a HTML man page and publishing it via `gh-pages`, but really that is way over the top given the size of the project (I may have done it just for the sake of interfacing with a web API like that but I've already done that for the [pull request mirror code](https://github.com/d-e-s-o/github-pull-request-mirror)).

My thinking is that ultimately the user should have a local man page anyway (when installing via a package) and that would take preference. The link from the README is mostly to get a quick impression of what is possible and I believe it serves that purpose just fine.

> Comments:
> - You could add a note about cargo fmt and clippy to CONTRIBUTING.md.
> - Did you consider moving CONTRIBUTING.md to the doc directory? As it’s mentioned in the README, interested developers should be able to find it anyway.

I'll address those two comments. Thanks!

Btw., I've also decided what I would like to release as `v0.2.0`. The main things I left out are the `pws` and the `lock` command. I'll include them in `v0.2.1`. They are fairly isolated, easy to integrate later, and not changing any existing interface. I believe it is beneficial to release a bit more often (and in smaller increments) to get a little more exposure: My other crates seemingly receive quite a traffic bump when publishing to `crates.io`. And if there are bug reports I don't have to release just to fix a single bug, but can include features as well. Don't feel pressured to package an updated version for Arch or Debian, though. I'll definitely ping you when `v0.2.1` is cut that is closer to being feature complete.

That being said, is there anything else that you see as a show stopper? If not I plan to finally push out a new version tomorrow.

nitrocli/README.md
d-e-s-o: Good catch, thanks!

robinkrahl:
As nitrocli has been accepted into Debian unstable, an HTML version of the man page for the latest release is available at [manpages.debian.org](https://manpages.debian.org/unstable/nitrocli/nitrocli.1.en.html).   I think the formatting is better than the PDF version.  The only drawback is that it does not contain the unreleased changes.  But as we expect users to run released versions, I think that’s fine.  What do you think about replacing the PDF with this link, or alternatively adding the link to the readme?

robinkrahl:
I don’t think that would makes sense to link one version of the man page in the Usage section and another version of the man page in the Packages section.  If you don’t want to change that paragraph, we shouldn’t link it at all.  (Also, I wouldn’t see this as necessarily Debian-specific – rather like docs.rs.  It’s an external service that automatically generates  an HTML version of the documentation for the most recent release.  It just happens to be based on the Debian package.)

d-e-s-o:
Hm, yeah. I guess the view I had in mind was: here is the link to the Debian package and this is the documentation for the very version packaged. For all I know the Debian package could be completely different from HEAD (which is what would be represented through the PDF), so in that sense I can see some value. But I also did not see a perfect fit. I don't have a better idea right now.

robinkrahl:
> here is the link to the Debian package and this is the documentation for the very version packaged.

Most of the time, there will be different versions in at least two of the three maintained suites, so ithat would be proably be rather confusing.

> For all I know the Debian package could be completely different from HEAD (which is what would be represented through the PDF)

I don’t think we *need* documentation that is up to date with HEAD.  Users should default to install a release via packages or via crates.io.  And anyone who compiles nitrocli from a Git clone should also be able to type `man ./doc/nitrocli.1`.

As the link I posted above points to the documentation for the unstable package, it should be up to date within one or two days after a release.

d-e-s-o:
> > For all I know the Debian package could be completely different from HEAD (which is what would be represented through the PDF)
> 
> I don’t think we _need_ documentation that is up to date with HEAD. Users should default to install a release via packages or via crates.io. And anyone who compiles nitrocli from a Git clone should also be able to type `man ./doc/nitrocli.1`.

`HEAD` may be the most common use case, but I think it's nice to be able to be able to navigate through the repository and get a matching readable documentation for any version currently being viewed.
It really depends who we want to cater to. When installing from packages that discussion is probably irrelevant because a man page is installed. As you said, for a Git clone the man page is there as well. I consider the installation from `crates.io` more as a preliminary hack to try out the program. Ultimately a package should be used. We don't have packages for many platforms, but that is a different problem in my opinion.

The main difference I see is between the view on Github (for which I want to keep things the way they are) and the README as published on `crates.io`. I could easily change the link before releasing (I do that anyway, because `crates.io` doesn't handle relative links properly), but I am not too eager to link to a largely Debian biased version of the man page. I have nothing against Debian (and I agree that the man page is nice to look at and the links to other man pages are great), I just feel that it should be neutral. Perhaps we could create a similar man page (without the branding) and then use https://linux.die.net/man/ ? I would link to that from `crates.io` but Github would be unaffected.

d-e-s-o:
We are up at https://www.mankier.com/1/nitrocli .

That's the only hoster I've heard back from so far. Publishing involves a pull request on [Github](https://github.com/jacksonp/ManKier-extra/). I noticed that `pinentry(1)` is not a proper link and I somehow think the bold text is a bit too heavy, but other than that it looks reasonable. I'll ask Jackson whether the overly long lines are intended to be that way. What do you think, @robinkrahl ?

robinkrahl:
> We are up at https://www.mankier.com/1/nitrocli .

Looks good! :-)

d-e-s-o:
For reference: the man page is now also available here: https://www.systutorials.com/docs/linux/man/1-nitrocli/

PR #44, opened 2019-01-05
Implement log level setting

d-e-s-o:
This patch set implements setting of the log level used by `libnitrokey` through a new command line option, -v/--verbose.


d-e-s-o:
@robinkrahl , mind giving it a review? :)

robinkrahl:
Looks good!  I’d only like to have more details in the man page:
- use at least `-vv` to see which commands are sent to the device
- use at least `-vvv` to see the full communication with the device
- `-vvvv` is the most verbose setting
- the logging output will be on stderr

d-e-s-o:
I'll add that. Thanks for the review!

d-e-s-o:
The change landed in `devel`.

PR #48, opened 2019-01-05
Add the --model option to restrict the device model to connect to

robinkrahl:
This patch adds the -m, --model option that can be used to restrict the
device model to connect to.  Per default, nitrocli connects to any
available Nitrokey device.  If this new option is set, it will instead
only connect to devices of the given Nitrokey model.

We introduce a new struct DeviceModel instead of using
nitrokey::DeviceModel to make sure that the command-line options are
parsed properly.  On the long term, we should add a connect_model
function to the nitrokey crate to make the connection code easier.


d-e-s-o:
Looks awesome, Robin. Thanks for implementing this feature! One comment: I think you forgot to add the new option to the command synopsis. I've taken care of that and recreated the PDF.

PR #51, opened 2019-01-06
Implement the first Nitrokey device test

d-e-s-o:
This pull request introduces the first device driven test for the program. It comprises three changes that enable us to effectively invoke the program's main function (that is in contrast to running `nitrocli` as part of a separate process; see the description of the second change for more details), supplying the desired arguments as well as buffers for stdout and stderr.

The fourth change adds the first actual test. We use the `nitrokey-test` crate to automatically serialize tests and run them only on present devices. (the test will probably not go in as-is, for there is a TODO that needs to be resolved first, but it shows the workings very well)

### DIFF
--- a/nitrocli/Cargo.lock
+++ b/nitrocli/Cargo.lock
@@ -56,6 +56,7 @@ dependencies = [
  "argparse 0.2.2",
  "libc 0.2.45",
  "nitrokey 0.3.0",
+ "nitrokey-test 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -79,6 +80,32 @@ name = "nitrokey-sys"
 version = "3.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 replace = "nitrokey-sys 3.4.1"
+
+[[package]]
+name = "nitrokey-test"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "proc-macro2 0.4.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 0.6.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.15.23 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "proc-macro2"
+version = "0.4.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unicode-xid 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "quote"
+version = "0.6.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "proc-macro2 0.4.24 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "rand"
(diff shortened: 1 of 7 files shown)


d-e-s-o:
May I ask for your eyes on the code, @robinkrahl ?

robinkrahl:
I haven’t tested the code yet, but I’d like to add some thoughts regarding the question of how to execute nitrocli.  We don’t want to test the libnitrokey functionality.  That’s already done in the `nitrokey` tests.  We want to test the `nitrocli` tool, and in my opinion, this includes running the executable the same way the user would run it.

The first `cargo run` might take some time, but given that most of the code has to be built for the tests anyway, this should only take a few seconds – one test case will probably take approximately the same time.

And we might not be able to perform type-safe error checks, but I don’t think we have to do that.  We don’t care whether a function returns `Error::CommandError(nitrokey::CommandError::InvalidSlot)` or `Error::Error("Invalid slot".to_string())`, we care about the exit status and a meaningful error message for the user.

But on the other hand, having customizable input and output channels adds complexity to the main code only for testing, and it can be a source of errors that are not detected by the tests (printing to stdout instead of the configured channel).

Are there more reasons against running the executable?

d-e-s-o:
Thanks for your comments, Robin!

> I haven’t tested the code yet, but I’d like to add some thoughts
> regarding the question of how to execute nitrocli. We don’t want to test
> the libnitrokey functionality. That’s already done in the nitrokey
> tests. We want to test the nitrocli tool, and in my opinion, this
> includes running the executable the same way the user would run it.

In my opinion you are painting a black/white scenario here which is not
accurate. No, we are not testing the binary, but we are testing the core
logic of it. Let's face it, all `nitrocli` does is provide a handy
command line interface for libnitrokey (perhaps with the potential of
some value add). I argue that with what I proposed we can certainly
cover all argument dispatching related aspects of the program (or at
least that it does not stand in the way of doing so).

Q: Is there a potential for unknowns?
A: Unqualified "yes".

The code layout will be different between the test and the final binary
and that could have side-effects on the program, for example. I argue
that the risk is manageable (if really deemed necessary we can add, say,
a Makefile level integration test that invokes the binary to test one
case) and not worth the price.
And even running `cargo` only gets you so far (sticking to the example I
provided): what about the difference between debug and release builds?
What about that between unstripped and stripped binaries? You will only
ever have so much confidence and it's not that we won't ever be running
`nitrocli` as part of the development.

> The first cargo run might take some time, but given that most of the
> code has to be built for the tests anyway, this should only take a few
> seconds – one test case will probably take approximately the same time.

Don't get me wrong, I don't particularly care about the test run time as
it stands (and that's in part because we don't have many tests and I can
only estimate how long they will take eventually). It was an example to
bring the point across that we are introducing unnecessary code into the
test path that we have absolutely no control over. It is also simply a
dependency that can be avoided.

> And we might not be able to perform type-safe error checks, but I don’t
> think we have to do that. We don’t care whether a function returns
> Error::CommandError(nitrokey::CommandError::InvalidSlot) or
> Error::Error("Invalid slot".to_string()), we care about the exit status
> and a meaningful error message for the user.

Agreed, and the beauty of the test infrastructure I propose is to not
rule out either :)
With a tiny bit of restructuring we can test at the level where all we
get is an int value (that in the `nitrocli` proper would be handed to
`exit`) and an stderr buffer, after the reported error was converted.

I am sure there are legitimate cases where we want to have more targeted
testing and it is beneficial to do that at a different level. Just
forking off a process does not allow you to do that. You have to live
with a black box, unconditionally. I argue that having an option is
good.

> But on the other hand, having customizable input and output channels
> adds complexity to the main code only for testing, and it can be a
> source of errors that are not detected by the tests (printing to stdout
> instead of the configured channel).

Yes, code needs to be written with testing in mind. So far we have not
done that, because testing was not a primary concern. And that in turn
certainly has been a concern of mine. But, just because there are some
changes in that direction does not mean that's bad: there was little
point to push for something we have little idea of what it will look
like in the end, and while I asked to move exiting of the program into
the main function and supply of stdout/stderr closer to it, that was as
much as I was sure about.
It's also not the case that this code is used *only* for testing. This
code is shared with the main program logic, that is exactly the point.

> (printing to stdout instead of the configured channel)

That's simple: a new feature will have to come with a test ;)

---

To move this request forward, how about the following proposal: I
massage the code a bit more so that tests can simply interface with
`main(stdout, stderr) -> i32`. This way we can test more similar to the
process level and, hence, have confidence into the error conversion
logic compared to what I have currently. Would that be okay?

robinkrahl:
I think we just have different opinions on the default way of testing a CLI application.  For me, spawning a new process with the command in question is the default way and so far, I don’t see any reason to prefer something else.  Of course both approaches have disadvantages and benefits, but I don’t see anything significant.  You see invoking cargo as an unnecessary complication, I refactoring the code to work with the tests.  So that’s a basic conflict that we won’t be able to solve (other than by chosing one ;)).

The code looks gode to me, so please merge it.

robinkrahl:
Regarding the suggested refactoring:  While I’d prefer to be as close to regular command execution as possible, we would lose some of the advantages you named, especially type-safe error matching.  If you think that’s worth it, go ahead, but you don’t have to.

d-e-s-o:
Thanks Robin. Yeah, I don't think we have to go "either or". You'll see what I mean once I have different tests done. I've pushed everything except for the test itself in slightly modified form.

PR #54, opened 2019-01-07
Add first set of integration tests

d-e-s-o:
This change introduces the first set of integration-style test for the
application. Those tests may or may not connect to an actual Nitrokey
device (depending on what they test). We use the nitrokey-test crate's
test attribute macro to automatically dispatch tests to connected
devices or skip them if a required device is not present. It also
provides the means for automatically serializing tests.

### DIFF
--- a/nitrocli/Cargo.lock
+++ b/nitrocli/Cargo.lock
@@ -1,4 +1,12 @@
 [[package]]
+name = "aho-corasick"
+version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 2.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "argparse"
 version = "0.2.2"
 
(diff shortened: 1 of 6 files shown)


d-e-s-o:
@robinkrahl , would be nice if you could take a look and let me know what you think. Do you think the test organization makes sense? I definitely think that we should not put everything into a single file (as we did for `commands.rs`). I envisioned splitting by command (only done for `status` right now), with `run.rs` containing some more or less generic tests pertaining the execution.

d-e-s-o:
If you feel that the patch is too big I can split it. I felt that a reasonable amount is boilerplate that does not need a close review.

robinkrahl:
I’ll have a look at it!  Regarding code organization, I think we could also consider splitting `commands.rs` into submodules (per subcommand).

d-e-s-o:
Thanks!

> Regarding code organization, I think we could also consider splitting `commands.rs` into submodules (per subcommand).

Yeah, I am looking into preparing a patch that removes some of the command enum related boilerplate (basically auto generating it). Let's see how bad it is after that. I believe there was a reason for the organization, but I don't recall exactly. I may just remember wrongly.

d-e-s-o:
Thank you :)

PR #55, opened 2019-01-07
Auto generate argument enums

d-e-s-o:
The argparse module we use for parsing events expects an enum in order
to convey what subcommand has been supplied as an argument. Such an enum
also needs to implement str::FromStr and, preferably, fmt::Display.
Manually writing down those definitions is error-prone, tedious, and
adds no value -- only lines of code.
As it turns out the proper definitions can be generated with relative
easy with a declarative macro, making the code much more concise. Hence,
with this change we use a newly introduced macro for generating those
enums.

### DIFF
--- a/nitrocli/src/arg_util.rs
+++ b/nitrocli/src/arg_util.rs
@@ -1,0 +1,79 @@
+// arg_util.rs
+
+// *************************************************************************
+// * Copyright (C) 2019 Daniel Mueller (deso@posteo.net)                   *
+// *                                                                       *
+// * This program is free software: you can redistribute it and/or modify  *
+// * it under the terms of the GNU General Public License as published by  *
+// * the Free Software Foundation, either version 3 of the License, or     *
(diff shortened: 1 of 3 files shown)


d-e-s-o:
Hey @robinkrahl , mind reviewing this change?

d-e-s-o:
Thanks for taking a look! Should not be a problem. I'll look at that as part of a subsequent change. Where I was actually going with that is (in case you haven't seen) that we can automatically enumerate all the `enum` variants programmatically: https://github.com/d-e-s-o/nitrocli/commit/7fa127200833cd79f4838a91b487df67e3297bf1

robinkrahl:
Looks familiar.  Didn’t you have that on the nitrokey-test branch?

d-e-s-o:
It's part of the extensions prototype where it helped me list all available extensions in the help text (along with all the built-in ones).

PR #58, opened 2019-01-10
Add support for non-interactive PIN entry

### LINKED ISSUE #27: Add non-interactive passphrase query
For testing (#14), we need a non-interactive way to enter passphrases.  I’d prefer to use environment variables to provide the passphrases.  Regardless of the mechanism, we could either compile it conditionally for tests, or add it is a feature for all users.  The downside of the latter option would be that it is much less secure than pinentry.  But I think we should let the user decide how to use the tool.  Even gpg provides a `--passphrase` option.  –  What do you think?

d-e-s-o:
This series of patches, at its core, addresses issue https://github.com/d-e-s-o/nitrocli/issues/27  by introducing environment variables that can be used to prevent any interactive pin entry. The tests make use of this functionality to be fully automated, even if no PINs are cached.

@robinkrahl Would you mind having a look at this one?

I plan to add at least two tests that would be interactive without this feature. I'll also update the man page, but haven't gotten around to doing so so far.

Closes #27


nitrocli/src/commands.rs
```diff
@@ -625,10 +625,30 @@ fn choose_pin(pin_type: pinentry::PinType) -> Result<String> {
   }
 }
 
+/// Choose a PIN of the given type.
+///
+/// If the user has set the respective environment variable for the
+/// given PIN type, it will be used.
```
robinkrahl: We cannot read the new PIN from the same environment variable as the old PIN.  (While `pin_unblock` uses the current admin PIN to set a new user PIN, the `pin_set` command requires an old and a new PIN of the same type.)  We should have `NITROCLI_{ADMIN,USER}_PIN_NEW` variables instead.

robinkrahl:
Commits one to four look good.

robinkrahl:
The man page update looks good.  I think I’d prefer bold variable names.  I had a look at some random man pages (**git**(1), **man**(1), **rustc**(1)) and all of them use boldface.  And I think you forgot the `pin unblock` command in the documentation for the `NITROCLI_USER_PIN_NEW` environment variable.

I agree with not mentioning the environment variables for every single command.  We could think about adding a short overview to the first section that also mentions the environment variables.  Regarding the required PINs, I think it would make sense to document that, especially for new Nitrokey users.

d-e-s-o:
> The man page update looks good. I think I’d prefer bold variable names. I had a look at some random man pages (**git**(1), **man**(1), **rustc**(1)) and all of them use boldface.

I'll make that adjustment. Thanks for taking a look!

> And I think you forgot the `pin unblock` command in the documentation for the `NITROCLI_USER_PIN_NEW` environment variable.

I did that intentionally, because that's how I implemented it. Conceptually, `unblock` should just do what it says, unblock the card by resetting the user PIN counter. I consider it sort of an implementation detail that it allows you to change the user PIN in interactive mode. Now obviously if an unblock is necessary the user either forgot the PIN or made a series of mistakes. However, it appears that setting the PIN he/she had in mind, i.e., `NITROCLI_USER_PIN`, makes perfect sense for those cases. Would you agree?

> I agree with not mentioning the environment variables for every single command. We could think about adding a short overview to the first section that also mentions the environment variables. Regarding the required PINs, I think it would make sense to document that, especially for new Nitrokey users.

Sounds good.

robinkrahl:
Ah, I only saw that you called `choose_pin` from `pin_unblock` and did not pay attention to the `new` parameter.  As both interactive `nitrocli` and `gpg` ask the user to choose a new PIN, I think it would consequent to use the `*_NEW` variable.  But I also see your point.  I think both are viable options; we just have to be clear about it in the documentation.

PR #59, opened 2019-01-10
Refactor get_device to use nitrokey::connect_model

robinkrahl:
nitrokey 0.3.1 introduced the connect_model function that connects to a
specific model given by an enum variant and returns a DeviceWrapper.
This new function allows us to remove the manual selection of a
connection method from the `get_device` function.  We only have to
implement From<DeviceModel> for nitrokey::Model to be able to convert
our model enum to nitrokey’s model enum.

### DIFF
--- a/nitrocli/src/args.rs
+++ b/nitrocli/src/args.rs
@@ -59,6 +59,15 @@ Enum! {DeviceModel, [
   Storage => "storage"
 ]}
 
+impl From<DeviceModel> for nitrokey::Model {
+  fn from(model: DeviceModel) -> nitrokey::Model {
+    match model {
+      DeviceModel::Pro => nitrokey::Model::Pro,
+      DeviceModel::Storage => nitrokey::Model::Storage,
+    }
+  }
+}
+
 /// A top-level command for nitrocli.
 Enum! {Command, [
   Config => "config",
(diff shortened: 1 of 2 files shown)


d-e-s-o:
> An unrelated question: What do you think about implementing `From<&str>` for `Error` to have `Error::from(s)` as a shorthand for `Error::Error(s.to_string())`? That would make the error handling code more concise.

Sounds good to me. I am all in favor of shortening where it makes sense.

d-e-s-o:
Change looks great, Robin. Thanks for remembering this code ;-) I've merged it.

One question: wouldn't it make more sense to have `connect_model` accept an `Into<DeviceModel>`? I feel like APIs that cross crate boundaries should be as flexible as possible with respect to arguments they accept.

d-e-s-o:
> > One question: wouldn't it make more sense to have `connect_model` accept an `Into<DeviceModel>`? I feel like APIs that cross crate boundaries should be as flexible as possible with respect to arguments they accept.
> 
> From an API design perspective, I don’t see what makes `Model` different from all other types (and we probably agree that we should not convert every type in the public interface into `Into<...>`). Now we do have a use case for `nitrocli` but I’m not sure whether we should change the API for this use case. I’ll think about it.

Yeah, I don't know. I am not saying it would be the right thing. I've read one or two Rust API guidelines but neither really touched on this. I believe the most important part is that there is no over specification. E.g., do not require a `String` when all you need is a `&str` or do not require a `&[T]` when all you want is an iterator. But that's obviously not the case here.

PR #60, opened 2019-01-11
Implement From<&str> for Error

robinkrahl:
This patch implements From<&str> for Error so that we can use
Error::from(s) as a shorthand for Error::Error(s.to_string()).  It
also replaces Error::Error with Error::from where possible.


d-e-s-o:
Merged. Thanks Robin!

PR #61, opened 2019-01-12
Improve multi-device support

d-e-s-o:
This change amends the logic we use for caching IDs to include the associated device's serial number. This way multiple devices will use different cache entries. It addresses https://github.com/d-e-s-o/nitrocli/issues/24.

@robinkrahl may I ask you to take a look?

Note that I did not include the model (pro/storage) in the cache ID as well. I am not sure if serial numbers are shared or not. But even if they are not, the chances of two IDs used by the same user having the same serial number is so small that I am tempted to just ignore this case in favor of a slightly simpler implementation.


robinkrahl:
Thanks, I’ll have a look at it!

I’d also like to add the serial number and the model to the pinentry description, so we’ll have to store the model in `PinEntry` anway and could add it to the cache ID.  But I agree that a collision is unlikely.

robinkrahl:
My only suggestion would be to add the model to the pinentry text (and possibly the cache ID).  We could add a `PinEntry::from(pin_type: PinType, device: &impl Device)` method that queries the serial number and the model of the device to reduce code complexity and duplication.

PR #63, opened 2019-01-13
Document supported platforms and macOS build

robinkrahl:


d-e-s-o:
Thanks for documentation that problem. I think I'd rather we frame that as a "limitation" or so. I tried to stay away from providing a list of supported platforms for the program as-is is pretty platform-independent and I have little control over what dependencies do and how that may change.

Perhaps something along the lines of:

#### Known Problems

Due to a problem with the default `hidapi` package on macOS, users are
advised to build and install [`libnitrokey`][] from source and then set the
`USE_SYSTEM_LIBNITROKEY` environment variable when building `nitrocli` using
one of the methods described above.

---

Is there an upstream issue (in the realm of the Nitrokey team) that we can reference?

robinkrahl:
> Is there an upstream issue (in the realm of the Nitrokey team) that we can reference?

None that I’m aware of – only Szczepan’s explanation in our own macOS
support ticket.

d-e-s-o:
Thank you Robin! Merged.

PR #64, opened 2019-01-14
Replace license identifier GPL-3.0+ with GPL-3.0-or-later

robinkrahl:
Cargo uses SPDX 2.1 license identifiers.  The identifier GPL-3.0+ is
deprecated as of version 2.0rc2 [0].  The current license identifier for GNU
General Public License v3.0 or later is GPL-3.0-or-later [1].

[0] https://spdx.org/licenses/GPL-3.0+.html
[1] https://spdx.org/licenses/GPL-3.0-or-later.html


d-e-s-o:
Thanks Robin! Merged.

PR #66, opened 2019-01-14
Add tests for otp command

d-e-s-o:
This change adds a set of tests for the otp command. We cover some
variants of the status, set, get, and clear. Testing all the possible
combinations is out of scope and so only a more or less arbitrary subset
of arguments was chosen.

@robinkrahl , interested in reviewing?

### DIFF
--- a/nitrocli/src/args.rs
+++ b/nitrocli/src/args.rs
@@ -463,7 +463,7 @@ fn otp_get(ctx: &mut ExecCtx<'_>, args: Vec<String>) -> Result<()> {
   let _ = parser.refer(&mut algorithm).add_option(
     &["-a", "--algorithm"],
     argparse::Store,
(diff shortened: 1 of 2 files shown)


robinkrahl:
Thanks!  Did you consider adding tests that validate the generated values?  I’d suggest to use an example from the test cases listed in the RFCs.

d-e-s-o:
Thanks for the review!

> Thanks! Did you consider adding tests that validate the generated values? I’d suggest to use an example from the test cases listed in the RFCs.

I did consider that. I was not sure whether it would add much value given that `nitrokey` already tests all that. But perhaps it's better than pretty much just testing for the length of a string. I'll add something.

robinkrahl:
Yeah, I think it would be good to have at least one test for HOTP and one test for TOTP.  Just to make sure that we put everything together correctly.

PR #68, opened 2019-01-15
Add the -V/--version option

robinkrahl:
The -V/--version option prints the nitrocli version to stdout and exits.
In the future, it should also print the used libnitrokey version, but as
the required function is only available with nitrokey 0.3.2 and as the
current interface does not reflect the latest change in version naming,
I skipped that in this patch.


d-e-s-o:
Merged. Thanks for taking this on, Robin!

d-e-s-o:
Ugh. It appears that `argparse`'s `Print` action does not honor the `stdout` `io::Write` argument passed to the parser. It just goes to `stdout` directly.

robinkrahl:
> Ugh. It appears that `argparse`'s `Print` action does not honor the
> `stdout` `io::Write` argument passed to the parser. It just goes to
> `stdout` directly.

Oh, I forgot to check that.  I think there’s no way around this for the
moment as every other action does not stop the argument processing.
I’ll prepare a patch for argparse.

robinkrahl:
I see you already did that.  Let’s hope it’ll be merged soon.

d-e-s-o:
> I see you already did that. Let’s hope it’ll be merged soon.

Doesn't look good. The author says that the crate is in maintenance-only mode, regards this pull request as a feature, and does not seem willing to merge.

PR #70, opened 2019-01-16
Auto-generate all the things!

d-e-s-o:
This patch set uses macro magic to generate more parts of the help texts. Unfortunately the `argparse` crate and its acceptance of references that must outlive the parser we work with makes it hard to make this code look nice. We have to introduce let-bindings for the texts of interest. `rustfmt` also adds some spice by completely reformatting some of the argument declarations.

Ideally we would want to somehow generate more of the parsing code and not just fractions of it. But I haven't found a good interface to hide all this boilerplate behind so far. I do realize that we are probably moving closer to what structopt would provide, but let's not get into that now :P

----

I need some unbiased eyes on these changes to tell me whether this is crazy or not. @robinkrahl , are you up for the job?

### DIFF
--- a/nitrocli/src/arg_util.rs
+++ b/nitrocli/src/arg_util.rs
@@ -109,8 +109,10 @@ macro_rules! Enum {
 /// replaced with a generated version of the enum's variants.
 macro_rules! fmt_enum {
   ( $enm:ident ) => {{
-    $enm
-      .all()
+    fmt_enum!($enm.all())
+  }};
+  ( $all:expr ) => {{
+    $all
       .iter()
       .map(::std::convert::AsRef::as_ref)
       .collect::<::std::vec::Vec<_>>()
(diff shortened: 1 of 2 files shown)


robinkrahl:
Wow, I’ll have a look at it tomorrow!

robinkrahl:
Very nice, thanks for implementing this!

Out of curiousity, why did you choose to combine the string building and formatting in `fmt_enum`?  My intuitive approach would have been to only return the string and let the caller, in this case `cmd_help`, take care of the formatting.

d-e-s-o:
Thanks for your review!

> Out of curiousity, why did you choose to combine the string building and formatting in `fmt_enum`? My intuitive approach would have been to only return the string and let the caller, in this case `cmd_help`, take care of the formatting.

That honestly came out of how I structured things initially. Basically I wanted to put more functionality into the macro, but I realized over time that this won't work as I had envisioned it. So I gradually moved functionality out again. It was only in the very last iteration that I decided to rename the macro to `fmt_enum`, because, well, that was what it had become. Truth be told, I asked myself the same question a while after finishing things. I may rework it slightly.

PR #72, opened 2019-01-20
Refactor handling of PINs

d-e-s-o:
Hidden volumes will not use a PIN but rather a password in order to be opened. The reasoning for this terminology is two-fold:
1) Nitrokey uses the term PIN for secrets that can only be retried a couple of times before a hardware lockout occurs, while passwords can be retried indefinitely.
2) As multiple hidden volumes can exist, each opened with a different password, we do not want to cache passwords. Hence, referring to passwords as opposed to PINs in this context makes it obvious that the `nitrocli pin clear` command will not have any effect.

This patch set makes the necessary refactorings to have PINs and passwords co-exist and be managed by the existing `pinentry` module (the name stays as it is in reference to the entity used for entering the PIN).

----

@robinkrahl, mind taking a look? Btw., I have the hidden volume support working and created my first volume 10 minutes ago! I still have to finish up the documentation and everything before review, though. I went with having a new subcommand, `storage hidden`, as you suggested. Pull request to land tomorrow.


robinkrahl:
Looks good!  Some suggestions:
- The examples for the firmware update password are eight characters long, so there might be a higher minimum password length.
- I’d let `cache_id` return an `Option<String>` and set the `"X"` in `inquire` – it is an implementation detail like escaping spaces with plus, not part of the API.
- Not related to your changes, but: Is the comment starting at line 238 still valid?  We *are* using the `--data` option.

d-e-s-o:
> Looks good! Some suggestions:

Thanks for your review!

> * The examples for the firmware update password are eight characters long, so there might be a higher minimum password length.

I suggest we check that when we implement support for that. Right now the new type is specific to hidden volumes (description etc.) and we probably want to add something similar to `PinType` if we require different passwords (or find another way to accomplish the same). 

> * I’d let `cache_id` return an `Option<String>` and set the `"X"` in `inquire` – it is an implementation detail like escaping spaces with plus, not part of the API.

Makes sense.

> * Not related to your changes, but: Is the comment starting at line 238 still valid?  We _are_ using the `--data` option.

I'll remove it. Thanks!

PR #74, opened 2019-01-23
Exclude CI scripts and rustmft config

robinkrahl:
The CI scripts and the rustfmt configuration are only needed when
developing.  There is no point in distributing them in the package
published on crates.io, so we exclude them from packaging using the
exclude setting in Cargo.toml.


d-e-s-o:
Merged. Thanks Robin!

PR #76, opened 2019-01-24
Document problems with running nitrokey-app

robinkrahl:
If nitrokey-app is running, the device it connected to can not be
detected by other applications.  This patch adds this issue to the list
of known problems in the readme.

### DIFF
--- a/nitrocli/README.md
+++ b/nitrocli/README.md
@@ -106,12 +106,16 @@ $ cargo build --release
 It is recommended that the resulting executable be installed in a
 directory accessible via the `PATH` environment variable.
 
-#### Known Problems
+Known Problems
+--------------
 
-Due to a problem with the default `hidapi` version on macOS, users are advised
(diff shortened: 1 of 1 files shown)


d-e-s-o:
I did notice that, too. Probably makes sense to document. Was that by any chance what @Dowwie is hitting?
Would it make sense to mention that `nitrokey-app` should be closed? It's pretty much obvious to me, but I am not sure whether that applies to everybody. For the other problem we provide a workaround/solution.

robinkrahl:
I don’t know, but it reminded me of this problem.  I’ll add some words on how to fix it.

d-e-s-o:
Thanks Robin! Merged.

szszszsz:
If I may, I think this is a good idea - suggesting potential issue causes during the tool's runtime should improve the UX.

d-e-s-o:
Yep, I think we are in agreement here.

PR #82, opened 2019-02-17
Respect ctx.std{out,err} for version output

robinkrahl:
Due to a bug in argparse [0], custom stdout and stderr settings are ignored
when using argparse::Print, as we currently do for the --version option. 
This patch adds a workaround for this problem:  Instead of using
argparse::Print, we use argparse::StoreTrue for the --version option.

The argument parsing will fail as the command is missing, but the version
variable will still be set to true if the version option was set.  So we
ignore the parsing result and discard the argparse output if the version
variable is set.

[0] https://github.com/tailhook/rust-argparse/pull/50


d-e-s-o:
Thanks for the pull request, Robin! I am not particularly fond of the idea of buffering all the output but as discussed via e-mail it seems to offer the best compromise (and it's a nifty idea :) ). I took the liberty to simplify the code a bit before merging.

PR #86, opened 2019-05-27
Allow for disabling of secret caching

d-e-s-o:
So far we have cached secrets in gpg-agent(1) whenever that made sense
to do (i.e., for the two PINs in most contexts but not for passwords).
While there is reason to believe that such caching is desired by the
majority of users, not everybody has a use for it.
To give users an opportunity to opt out of such caching, this change
introduces a new environment variable, NITROCLI_NO_CACHE, that, when
present in the environment, instructs the program to bypass the cache
for all operations that require a secret and to instead inquire such
secrets each time they are needed.


d-e-s-o:
@robinkrahl , are you okay with this approach? Do you see problems with the code? If not, and if Andrew confirms that this change works for him, I would want to merge it and then tag another release (it's been a while and a few new things have accumulated). This would presumably be the last release on the 0.2.x branch.

My plan is to get the storage refactoring into 0.3.

robinkrahl:
Sorry for the late review – looks good to me.

PR #88, opened 2019-05-27
Remove deprecated --ascii option of otp set subcommand

d-e-s-o:
The --ascii option of the otp set subcommand has been deprecated a while
in favor of --format. As the next release is slated to be a major one
breaking backwards compatibility, this change removes this option for
good.


d-e-s-o:
Should be a quick and easy one. @robinkrahl if you find the time, can you please take a look? Is there anything else that we deprecated and would be able to remove?

robinkrahl:
Looks good (although I did not yet test it)!  I can’t think of anything else right now.

PR #89, opened 2019-06-08
Add support for changing read-write mode of unencrypted volume

d-e-s-o:
This change adds support for changing the read-write mode of the
unencrypted volume. To do so, we introduce a new top-level command,
unencrypted, with a new subcommand, set, that accepts the new mode of
the volume.


d-e-s-o:
@robinkrahl would you mind taking a look at the change?

robinkrahl:
Nice, thanks for implementing this!

d-e-s-o:
Thanks Robin!

PR #96, opened 2020-01-04
Update packaging instruction for Debian and Ubuntu

robinkrahl:
This patch updates the packaging instruction for Debian:  Instead of
updating the TODO.rst file, we now have to create a RFS file in the
source directory.

This patch also adds links to the trackers for the Debian and Ubuntu
packages.


d-e-s-o:
Merged as https://github.com/d-e-s-o/nitrocli/commit/203e691f46d591a2cc8acdfd850fa9f5b0fb8a98. Thanks Robin!

Issue #92, opened 2019-08-25
Use different crate for argument parsing


d-e-s-o:
The author of `argparse` has made it fairly clear that the crate is in maintenance mode (at best; at worst I'd say it's abandoned). While it does pretty much what we want it to, there are flaws that I find increasingly annoying:
1) We had to implement some workaround for a bug in its code that won't get fixed because the proposed fix broke backwards compatibility.
2) The help texts it builds and displays are extremely rigid in their layout, providing a sub-par user experience (I've always dislike how we list all available commands/subcommand only on a single line) and not playing nice with extensions, of which there could be a lot.
3) Let's face it, the way it is used is awkward and definitely not as obvious as it could be. While our code base has evolved around it fairly nicely, I still find myself trying to understand nuances related to argument handling over and over again.

So at this point I would be interested in considering alternatives. The only viable one I am aware of is `structopt`. In the past we've shied away from it due to bloat (without having conducted actual testing on *this* crate), and while that is still a concern, I'd say point 2) above tipped me over as user experience over all is still higher ranked.

We should start by prototyping use of `structopt` and having a look at the outcome (code structure and resulting binary size).

d-e-s-o:
Not to forget that [bash completion](https://github.com/d-e-s-o/nitrocli/issues/57) *should* come much easier when we use this crate.

robinkrahl:
Please have a look at my [`clap` branch][] which contains a first implementation using `structopt`.  There are some issues with it:
- We still have to implement the error handling (i. e. choose the correct input and output streams, determine the return value).  Currently I just call [`clap::Error::exit`][] (see [`args.rs:handle_arguments`][]).
- Currently, the available options for an argument are not displayed in the help message.  We might have to do that manually.

But the code is a lot shorter and a lot more readable – and it should get even better if we improve the `Command!` macro (→ no type annotation in the closure, no need for empty structs).

Here are some measurements of the code and binary size:

```
$ git checkout -q devel && wc nitrocli/src/arg*.rs
  984  3205 28739 nitrocli/src/args.rs
  158   562  4498 nitrocli/src/arg_util.rs
 1142  3767 33237 total
$ git checkout -q clap && wc nitrocli/src/arg*.rs
  540  1872 16250 nitrocli/src/args.rs
  133   434  3674 nitrocli/src/arg_util.rs
  673  2306 19924 total
[...]

PR #97, opened 2020-01-08
Replace argparse with structopt

robinkrahl:
As discussed in #92, this patch series reworks the argument handling and
replaces argparse with structopt.


nitrocli/src/arg_util.rs
```diff
@@ -107,14 +107,14 @@ macro_rules! enum_int {
...
-      type Err = ();
+      type Err = &'static str;
 
       fn from_str(s: &str) -> ::std::result::Result<Self, Self::Err> {
         match s {
           $(
             $str => Ok($name::$var),
           )*
-          _ => Err(()),
+          _ => Err("[error]"),
```
d-e-s-o: Can this be a meaningful error? Can we include the string and the available variants? Or is it not user-visible? (looks user-visible to me)

You could use something like this, I suppose:
```
    $all
      .iter()
      .map(::std::convert::AsRef::as_ref)
      .collect::<::std::vec::Vec<_>>()
      .join(",")
```

nitrocli/src/arg_util.rs
```diff
@@ -25,6 +25,58 @@ macro_rules! count {
 }
 
 macro_rules! Command {
+  ( $name:ident, [ $( $var:ident($inner:ident) => ($str:expr, $exec:expr), ) *] ) => {
```
d-e-s-o: Hm, is this a huge code duplication? But I can't seem to find it in the end result, so perhaps it's okay?

nitrocli/src/arg_util.rs
```diff
@@ -76,6 +76,8 @@ macro_rules! Command {
         }
       }
     }
+
+    impl_default!($name => $( $name::$var(::std::default::Default::default()) , )*);
```
d-e-s-o: Seems to me a bit like a misuse of `Default`. Or at least a "dangerous" interpretation of it. But I guess it's only temporary as well?

nitrocli/src/args.rs
```diff
@@ -129,21 +143,271 @@ impl From<DeviceModel> for nitrokey::Model {
...
+/// Interacts with the device’s encrypted volume
+#[derive(Debug, Default, PartialEq, structopt::StructOpt)]
+pub struct EncryptedArgs {
+  #[structopt(subcommand)]
+  subcmd: EncryptedCommand,
+}
+
+/// Closes the encrypted volume on a Nitrokey Storage
+#[derive(Debug, Default, PartialEq, structopt::StructOpt)]
+pub struct EncryptedCloseArgs {}
```
d-e-s-o: Would you mind merging in my patch such that you can remove those empty structs? Alternatively just adding it to the list of commits works, too, if it's too much work.

I think one thing lacking is that we do not currently handle meta information in our macros (e.g., doc attributes). I think you can match on
```rust
#[$docs:meta])* 
```
to accomplish that.

nitrocli/src/args.rs
```diff
@@ -129,21 +143,271 @@ impl From<DeviceModel> for nitrokey::Model {
...
+
+/// Opens the encrypted volume on a Nitrokey Storage
+#[derive(Debug, Default, PartialEq, structopt::StructOpt)]
+pub struct EncryptedOpenArgs {}
+
+/// Interacts with the device’s hidden volume
```
d-e-s-o: Would it be possible to agree on using ' over ’ as apostrophe? I know you always use the latter, but I argue ' is more prevalent in the code base (and I don't even know how to produce ’ to be honest...)

nitrocli/src/args.rs
```diff
@@ -129,21 +143,271 @@ impl From<DeviceModel> for nitrokey::Model {
...
+  /// Shows the name stored on the slot
+  #[structopt(short, long)]
+  name: bool,
+  /// Shows the login stored on the slot
+  #[structopt(short, long)]
+  login: bool,
+  /// Shows the password stored on the slot
+  #[structopt(short, long)]
+  password: bool,
+  /// Prints the sotred data without description
```
d-e-s-o: sorted

d-e-s-o:
I've started going through your changes. Very nice stuff! Thanks for splitting things into reasonable chunks while keeping everything compilable. I look forward to the updated version. By any chance, can you fix the test failures while at it? :-)

nitrocli/src/arg_util.rs
robinkrahl: Good catch, I just forgot about it.

nitrocli/src/arg_util.rs
robinkrahl: (1) Only the `Display` implementation is duplicate – the other implementations are different (`Command::Variant` vs `Command::Variant(_)`). (2) The old macro variant is deleted after the migration.  In the first implementation, I used internal macros, but that just complicates the transition.

nitrocli/src/arg_util.rs
robinkrahl: I think it’s an acceptable implementation, but it’s only temporary anyway.  For structopt, we have to construct an instance of the subcommand enums.  That would be very complicated without `Default`.  The implementation and the macro are removed in the “Remove unused code after structopt transition“ commit.

nitrocli/src/args.rs
robinkrahl: I don’t care that much about it, but I can’t promise that I can change my muscle memory. ;)  It is Alt Gr + Shift + N on X11, btw.

robinkrahl:
The failure of the version test is caused by `clap` sending some stuff directly to `stdout`, see https://github.com/clap-rs/clap/issues/1390. I set it to ignore.

robinkrahl:
Thanks for the review! I fixed all issues except the two temporary ones and updated the pull request.

robinkrahl:
Regarding https://github.com/d-e-s-o/nitrocli/commit/c0d8583e0aba69b0628eb9c036f50a86f33d381f – to add support for generating completion scripts, which I intend to do after this is merged, we need to access the `Args` struct from `build.rs`. So it would be easier to just keep it public for the time being.

d-e-s-o:
Merged as https://github.com/d-e-s-o/nitrocli/commit/6c31f18cf40c7662b34ee89b3e6ddf516730611e with some minor changes. Thanks again for your work on this feature!

robinkrahl:
Thanks!

Issue #100, opened 2020-01-22
Configuration file


robinkrahl:
What do you think about having defaults for at least some of the arguments in a configuration file?  For example, once we have a `--serial-number` option, I’d like to be able to set the serial number of my production device for all `nitrocli` calls.

An implementation could use `config` for the file parsing and `app_dirs` to get XDG-compliant paths.  (This would pull in `serde` and `nom`.)

d-e-s-o:
The thought has crossed my mind. It's probably a good idea. I hope it can be implemented easily such that command line arguments take precedence over what is in the config. Things may get tricky with the `conflicts_with` attributes that `structopt` supports and we use (we may have to find a better way for those; I recently stumbled over [this issue](https://github.com/TeXitoi/structopt/issues/320), which may be of interest in this context). Perhaps we can layer this functionality before argument parsing (it's what I've done elsewhere and it has served me well so far)? Just a thought.

I am not familiar with `config` and only had a cursory glance. I suspect we will also need one of the backends, presumably `toml` (or `yaml`?). I am actually not quite sure what value-add the crate provides over just using `serde` with `toml` and [a custom derive](https://docs.rs/toml/0.5.6/toml/#deserialization-and-serialization). Well, I sort of see it, just not really the benefit for our context. Have you looked closer already by any chance?

robinkrahl:
> I hope it can be implemented easily such that command line arguments
> take precedence over what is in the config.

Yes, I agree.  We should prefer arguments over environment variables
over user configuration over system configuration.

> Perhaps we can layer this functionality before argument parsing (it's
> what I've done elsewhere and it has served me well so far)?

Yes, I agree.  I thought of adding the data to the {Run,Exec}Ctx,
similar to the environment variables.

> I suspect we will also need one of the backends, presumably `toml` (or
> `yaml`?).

I’d prefer toml.  It is simple enough for non-Rust users to get used to,
and it is quite similar the common INI-style configuration files.

> I am actually not quite sure what value-add the crate provides over just
> using `serde` with `toml` and [a custom
> derive](https://docs.rs/toml/0.5.6/toml/#deserialization-and-serialization).

It would make it easier to merge multiple configuration sources (e. g.
system configuration in /etc/xdg/nitrocli/config(.toml?), user
configuration in ~/.config/nitrocli/config(.toml?) and environment
variables).

I think the major decision is which arguments to support.  We could
support all arguments, with subgroups for subcommands, e. g.:

[...]

PR #101, opened 2020-01-25
Parse configuration file

robinkrahl:
This patch series reads selected options (currently no_cache and model) from a
configuration file.  This is only for discussion in #100.  As I already
implemented it, I thought I might as well share it, even if we end up using a
different approach.


d-e-s-o:
Just to make sure, there is nothing you are waiting on from me here, correct @robinkrahl ?

robinkrahl:
I’m waiting on a decision on #100 – if you want to go with this
approach, I can update the pull request with the missing documentation
etc.

robinkrahl:
I updated the pull request – I think it could be merged now.

I made these changes to the patch series:
- Replace `app_dirs` with `directories`.  `app_dirs` has not been updated for a while, and the author did not respond to my question regarding the maintenance status.  By switching to `directories`, we lose access to system-wide configuration files, but we gain a more robust and better maintained dependency.  (Problems with `app_dirs` include that it tries to create directories even if told not to, see AndyBarron/app-dirs-rs/issues/34, and even if it supports system-wide configuration, it does not comply with the specification, see AndyBarron/app-dirs-rs/issues/12).
- Squash and re-order the commits to make the patch series more concise and to avoid unnecessary intermediate steps.
- Add a `--no-cache` option for consistency.
- Update the documentation and add an example file.
- Add a simple unit test.

d-e-s-o:
Thank you, Robin. I hope to take a closer look tomorrow. I'll probably merge it if it looks alright.  From my perspective, it mostly comes down to technical details that differentiate the solutions and I don't have the time to experiment with alternative approaches right now.

d-e-s-o:
Seems as if we are pulling in two version of `arrayvec` and now have a total of three mechanisms to deduce/do something based on the rustc version we work with.

robinkrahl:
Sorry, I did not mean to close this, I was just cleaning up old branches.

> Seems as if we are pulling in two version of `arrayvec`

Ultimately, `nom` is to blame here as they use an old version of `lexical-core`, see Geal/nom/pull/1100 and Geal/nom/pull/1101.  (The pull request would not fix this as it updates `lexical-core` to 0.6, but the `arrayvec` dependency is updated in 0.7.)

> now have a total of three mechanisms to deduce/do something based on the rustc version we work with.

As far as I see, they are similar but not the same:
- `rustc_version` provides functions that are evaluated at runtime.
- `rustversion` provides macros for conditional compilation.

Which one is the third?

d-e-s-o:
> > Seems as if we are pulling in two version of `arrayvec`
> 
> Ultimately, `nom` is to blame here as they use an old version of `lexical-core`, see [Geal/nom/pull/1100](https://github.com/Geal/nom/issues/1100) and [Geal/nom/pull/1101](https://github.com/Geal/nom/pull/1101). (The pull request would not fix this as it updates `lexical-core` to 0.6, but the `arrayvec` dependency is updated in 0.7.)

Thanks for digging. I understand, but irrespective of who's fault it is, it's still not nice and something I'd really want to avoid.

> > now have a total of three mechanisms to deduce/do something based on the rustc version we work with.
> 
> As far as I see, they are similar but not the same:
> 
> * `rustc_version` provides functions that are evaluated at runtime.
> * `rustversion` provides macros for conditional compilation.
> 
> Which one is the third?

`version_check` is what I was referring to.

d-e-s-o:
Just noticing the update. Is this the updated pull request ready for review already, Robin?

robinkrahl:
No, I just rebased it onto the current devel.  I’ll open a new PR with
the new implementation so that you can compare both, okay?

d-e-s-o:
> No, I just rebased it onto the current devel. I’ll open a new PR with the new implementation so that you can compare both, okay?

Sounds good.

d-e-s-o:
We can probably close this pull request now that v2 has been merged.

PR #102, opened 2020-01-26
Make --verbose and --model options global

robinkrahl:
This patch adds the attribute `global = true` for the top-level
--verbose and --model options, which ensures that they can also be set
for subcommands.  For example:
    nitrocli status --model pro
Instead of only:
    nitrocli --model pro status

### DIFF
--- a/nitrocli/CHANGELOG.md
+++ b/nitrocli/CHANGELOG.md
@@ -5,6 +5,7 @@ Unreleased
   - Added `structopt` dependency in version `0.3.7`
   - Replaced `argparse` with `structopt`
   - Removed the `argparse` dependency
+  - Made the --verbose and --model options global
 - Bumped `nitrokey` dependency to `0.5.1`
 
 
--- a/nitrocli/src/args.rs
+++ b/nitrocli/src/args.rs
@@ -74,10 +74,10 @@ impl<'io> Stdio for ExecCtx<'io> {
 #[structopt(name = "nitrocli")]
 struct Args {
   /// Increases the log level (can be supplied multiple times)
-  #[structopt(short, long, parse(from_occurrences))]
+  #[structopt(short, long, global = true, parse(from_occurrences))]
   verbose: u8,
   /// Selects the device model to connect to
-  #[structopt(short, long, possible_values = &DeviceModel::all_str())]
+  #[structopt(short, long, global = true, possible_values = &DeviceModel::all_str())]
   model: Option<DeviceModel>,
   #[structopt(subcommand)]
   cmd: Command,


d-e-s-o:
Nice! I wasn't aware of this option. Very handy. I've [merged](https://github.com/d-e-s-o/nitrocli/commit/262df06856c02de25486afebe2956d0dee07356c) the change.

Are you aware of a way to express this new capability in the `man` page? Not that it's a big deal...

robinkrahl:
> I wasn't aware of this option.

Me neither.  I found it when I read the clap documentation.

> I've merged the change.

Thanks!

> Are you aware of a way to express this new capability in the `man` page? Not that it's a big deal...

I thought about that too, but decided not to change anything.  Maybe we
could change this line:

> nitrocli [-m|--model pro|storage] [-v|--verbose] command [arguments]

to:

> nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]

Or we don’t mention the global attributes at all in the synopsis line.

d-e-s-o:
> nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]

Where would the `--version` be in that version version of the synopsis? Perhaps we could add a second line to it:

```
nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]

nitrocli [--help] [--version]
```
(as seen in `xmlto(1)`)

> Or we don’t mention the global attributes at all in the synopsis line.

Perhaps if we introduce a "GLOBAL OPTIONS" (or "FLAGS", or so) section, to differentiate `--version` from `--verbose`. Otherwise I find it more confusing. Hm.

I guess I may be coming to the same conclusion as you did.

robinkrahl:
> > nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]
> 
> Where would the `--version` be in that version version of the synopsis?

I accidently looked at the manpage installed on my system which does not
yet have the --version option.  I agree that --version is problematic
with this approach.

> Perhaps we could add a second line to it:
> 
> ```
> nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]
> 
> nitrocli [--help] [--version]
> ```
> (as seen in `xmlto(1)`)

I think I intended to do that at one point, but then decided against it
for some reason that I cannot remember.  I might have based my decision
on the Git man pages which I like to use as a reference.

If we do something like this, I’d rather list each exclusive option
separately:

> nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments]
> nitrocli --help
> nitrocli --version


> > Or we don’t mention the global attributes at all in the synopsis line.
>
> Perhaps if we introduce a "GLOBAL OPTIONS" (or "FLAGS", or so)
> section, to differentiate `--version` from `--verbose`. Otherwise I
> find it more confusing. Hm.

That’s true.  But what term would you use to describe --version?  It’s
an option or flag that behaves like a command.  We could add a version
command and list --version together with the command in the COMMANDS
section:

    COMMANDS
        General
	    nitrocli version
	    nitrocli --version
	        Print the nitrocli version and exit.

	    nitrocli help
	    nitrocli --help
	        Print usage information for nitrocli and exit.
	
	    nitrocli list [-n|--no-connect]
	    	and so on

But that would probably be even more confusing.

Anyway, removing the options from the synopsis would make it easier to
read if we add both --serial-number and --usb-path.

d-e-s-o:
Sorry for the late response. This thread got somewhat lost..

>> If we do something like this, I’d rather list each exclusive option separately:
>> nitrocli command [-m|--model pro|storage] [-v|--verbose] [arguments] nitrocli --help nitrocli --version


> Or we don’t mention the global attributes at all in the synopsis line.

Yes, if that helps shorten things in light of the new upcoming options I am fine with that as well.

>> Perhaps if we introduce a "GLOBAL OPTIONS" (or "FLAGS", or so) section, to differentiate `--version` from `--verbose`. Otherwise I find it more confusing. Hm.

> That’s true. But what term would you use to describe --version? It’s an option or flag that behaves like a command.

I guess the "global" would be the differentiating factor.

> Anyway, removing the options from the synopsis would make it easier to read if we add both --serial-number and --usb-path.

Sounds good to me.

PR #103, opened 2020-01-29
doc: Remove options from synopsis in man page

robinkrahl:
As it is no longer required to set the global options before the command
and as we will probably introduce more global options with the next
releases, this patch removes the global options from the synopsis line
in the man page.  See issue #102 for more details.


d-e-s-o:
Thank you!

PR #104, opened 2020-02-03
Update nitrokey dependency to 0.6.0

robinkrahl:
nitrokey 0.6.0 introduced the SerialNumber struct (instead of
representing serial numbers as strings).  We no longer have to manually
format the serial number as SerialNumber implements Display.

For consistency, we keep the old formatting for the cache ID used by
pinentry.
TODO: should we use the new format instead?

### DIFF
--- a/nitrocli/src/pinentry.rs
+++ b/nitrocli/src/pinentry.rs
@@ -54,7 +54,7 @@ pub trait SecretEntry: fmt::Debug {
 pub struct PinEntry {
   pin_type: PinType,
   model: nitrokey::Model,
-  serial: String,
+  serial: nitrokey::SerialNumber,
 }
 
 impl PinEntry {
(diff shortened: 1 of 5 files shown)


nitrocli/src/pinentry.rs
```diff
@@ -79,7 +79,8 @@ impl PinEntry {
 impl SecretEntry for PinEntry {
   fn cache_id(&self) -> Option<CowStr> {
     let model = self.model.to_string().to_lowercase();
-    let suffix = format!("{}:{}", model, self.serial);
+    // TODO: directly format serial instead?
+    let suffix = format!("{}:{:08x}", model, self.serial.as_u32());
```
robinkrahl: Should we use the new format with the `0x` prefix, or should we stick with the old format without prefix?

d-e-s-o: I was going to look what we display on our `pinentry` dialog, but it seems you already did that.
> For consistency, we keep the old formatting for the cache ID used by
pinentry.

I have a slight preference for preserving the existing behavior.

d-e-s-o: Actually, nevermind what I said. Given that we already use the other format in other places anyway I am okay with switching over. I'll take care of that.

d-e-s-o:
Merged as https://github.com/d-e-s-o/nitrocli/commit/51d0fbb73eb42325fb2a0832810fd9e1d4339743. Thanks for the change, Robin!

PR #107, opened 2020-04-05
Add support for generating a bash completion script

d-e-s-o:
This pull request adds support for generating a bash completion script. If
sourced, the shell will provide tab completions for the program's
arguments.
There are two possible approaches provided by clap for going about
generating shell completion functionality: either at build time by
separately generating the clap parsers out-of-band or at run time, as an
option to the main program itself. We are generally not too much in
favor of a run time approach, as it means less inspectability at
installation time and more overhead in the form of code crammed into the
main binary.
Hence, with this change we take the "build time" approach. Clap
recommends hooking the generation up in build.rs, but this seems like an
inflexible choice. For one, that is because it would mean
unconditionally generating this file or some user-unfriendly environment
variable based approach to making the process conditional. But also
because specifying the command for which to generate the script should
likely be configurable. That is a limitation of the completion script
that clap generates (see https://github.com/clap-rs/clap/issues/1764).
Instead, we provide a utility program that emits the completion script
to standard output, accepting regular command line options itself. In
so doing we allow for installation time generation of the completion
script or installation of the utility itself, the output of which could
be sourced on demand -- depending on the user's preference.

---

@robinkrahl If you have the time, can you please take a look? I think we could now think about consolidating `args.rs` and `commands.rs`. Not sure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Unreleased
   - Replaced `argparse` with `structopt`
   - Removed `argparse` dependency
   - Made the --verbose and --model options global
+- Added bash completion support via `shell-complete` utility program
 - Removed vendored dependencies and moved source code into repository
   root
 - Bumped `nitrokey` dependency to `0.6.0`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,10 @@ exclude = ["ci/*", "rustfmt.toml"]
 [badges]
 gitlab = { repository = "d-e-s-o/nitrocli", branch = "master" }
 
+[[bin]]
+name = "shell-complete"
+path = "var/shell-complete.rs"
+
 [profile.release]
 opt-level = "z"
 lto = true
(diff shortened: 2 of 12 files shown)


robinkrahl:
Nice, thank you!

Some minor suggestions:
- `shell-complete --help` should mention that it prints the completion script to stdout.
- I’d prefer a more specific binary name, for example `nitrocli-shell-complete`.

I would also support merging `args.rs` into `commands.rs` (maybe except for `handle_arguments` which could also go into `main.rs`). `arg_defs.rs` could then become `args.rs`.

d-e-s-o:
Thanks for your review!

> * `shell-complete --help` should mention that it prints the completion script to stdout.

Done.

> * I’d prefer a more specific binary name, for example `nitrocli-shell-complete`.

I'd rather not prefix something in the nitrocli repository with nitrocli itself. It seems like useless repetition. I'd say that if users/distributors require a more unique name, they can always rename at installation time.

> I would also support merging `args.rs` into `commands.rs` (maybe except for `handle_arguments` which could also go into `main.rs`). `arg_defs.rs` could then become `args.rs`.

Sounds good. Let me do that as a follow up, though.

PR #110, opened 2020-08-05
Change default OTP format to base32

d-e-s-o:
An arguably unrepresentative survey of services (GitHub, Google
Authenticator, and Bitbucket) seems to suggests that the base32 format
is the de-facto standard format for OTP secrets. Given that it's not
necessarily obvious what format a secret is in and that most services
refrain from mentioning it explicitly, having the correct default format
is fairly important.
With this change we switch the default format from hexadecimal to
base32 to accommodate for this finding.


d-e-s-o:
@robinkrahl can you take a quick look? (not sure why I can't add you as a reviewer myself...)

robinkrahl:
Looks good to me.

To add me as a reviewer, you’d have to explicitly grant me read access
to the repository according to the documentation:
	https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review

PR #113, opened 2020-08-31
Topic/anyhow

d-e-s-o:


d-e-s-o:
Here is the patch set for moving to `anyhow` for error handling. @robinkrahl, may I ask you to take a look? I guess mostly high level. I still haven't given it the final polish (or looked at the test/lint failures in detail), but will do so before check in.

I'd actually want to merge this work next and only then come to the config work you have done (which I understand also is not quite ready because of a crate not yet being published), mostly because I'd want to avoid rebasing all this `.context(...)` stuff and reevaluating call sites again, as I had already gone over all of that. I can take care of rebasing your pull request and making it work a bit more nicely with the changed error handling model. Would that be okay with you?

Cargo.lock
```diff
@@ -9,6 +9,12 @@ dependencies = [
  "memchr",
 ]
 
+[[package]]
+name = "anyhow"
+version = "1.0.28"
```
robinkrahl, line 14: Could be updated to 1.0.32.

src/args.rs
```diff
@@ -134,13 +134,13 @@ pub enum ConfigOption<T> {
 }
 
 impl<T> ConfigOption<T> {
-  pub fn try_from(disable: bool, value: Option<T>, name: &'static str) -> Result<Self, String> {
+  pub fn try_from(disable: bool, value: Option<T>, name: &'static str) -> anyhow::Result<Self> {
     if disable {
       if value.is_some() {
-        Err(format!(
+        anyhow::bail!(
```
robinkrahl, line 140: `anyhow::ensure!(value.is_none(), ...)` might be easier to read.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 - Changed default OTP format from `hex` to `base32`
+- Updated minimum required Rust version to `1.42.0`
+- Bumped `nitrokey` dependency to `0.7.1`
```
robinkrahl, line 5: List the anyhow dependency and the error refactoring?

robinkrahl (approved):
LGTM.  Thank you!

d-e-s-o:
Thanks for your review, Robin! Everything is in `devel` now.

PR #115, opened 2020-09-01
Topic/copyright

d-e-s-o:


d-e-s-o:
@robinkrahl as we discussed in your recent pull request, here is the change to move over to complying with REUSE. Please take a look when you get a chance.

robinkrahl:
Great, thank you!  For full compliance with the specification, we would
have to add copyright annotations to all files (see `reuse lint` [0]),
but this is already a big improvement.

[0] https://github.com/fsfe/reuse-tool

d-e-s-o:
Thanks for your review! I've added this information to all files, added a CI pass, and merged the change.

robinkrahl:
Thanks for the update!  You’re probably already know that, but I just wanted to note that you can use HTML comments in markdown files instead of separate license files.

PR #116, opened 2020-09-03
Ctx

robinkrahl:
This PR combines the identical structs `RunCtx` and `ExecCtx` into `RunCtx`.  (We could even drop the prefix and just use `Ctx` instead.)


robinkrahl:
True, that’s even better. I’ve updated the PR.

d-e-s-o:
All in `devel` now! Thanks! I also removed the `Stdio` trait, which has not been in use for a while.

PR #117, opened 2020-09-03
Declare semver compatibility and public API

robinkrahl:
This patch adds a new section to the readme that declares that we follow
the Semantic Versioning specification and that the public API of or
crate is defined by the man page.

----

In #112, we talked about the issue of stability and defining nitrocli’s public API.  This patch attempts to solve this issue by declaring the nitrocli man page as nitrocli’s public API.  I think this is a quite elegant solution as it allows us to change implementation details that are not mentioned in the man page (e. g. output formatting) while providing a stable interface to for using nitrocli in scripts and extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,13 @@ Known Problems
   running shortly afterwards ([upstream issue][libnitrokey#137]).
 
 
+Public API and Stability
+------------------------
+
+**nitrocli** follows the [Semantic Versioning specification 2.0.0][semver].
+Its public API is defined by the [nitrocli(1) `man` page](doc/nitrocli.1.pdf).
+
+
 Contributing
 ------------
 
@@ -179,3 +186,4 @@ the full text of the license.
 [gplv3-tldr]: https://tldrlegal.com/license/gnu-general-public-license-v3-(gpl-3)
 [libnitrokey#32]: https://github.com/Nitrokey/libnitrokey/issues/32
 [libnitrokey#137]: https://github.com/Nitrokey/libnitrokey/issues/137
+[semver]: https://semver.org


d-e-s-o:
Thanks for the pull request! I actually did not mean my comment to be a call to action ;-) I am admittedly not super eager to make such a compatibility statement. At least not at this point. Consider for instance the change of the default in the OTP format that we made a while back (147d01663f9b95c6819d3c6afe8694cd291dbc3d). According to the definition you propose it would have been a breaking change, wouldn't it? It just seems overly sensitive to say so.

But then if we really follow semver.org non of what we are doing right now matters anyway, until we are 1.0, isn't that right?

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Or did I miss anything? Is that what you intended?

For that reason I actually like Cargo's model much more, [which states](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements):
> While SemVer says there is no compatibility before 1.0.0, Cargo considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.

That would at least have a meaning to users right now.

robinkrahl:
> But then if we really follow semver.org non of what we are doing right
> now matters anyway, until we are 1.0, isn't that right?

Exactly.  The goal of this PR is to have a solution for the 1.0 version,
which I feel is coming closer as extension support is the only remaining
major issue.

> Consider for instance the change of the default in the OTP format that
> we made a while back (147d01663f9b95c6819d3c6afe8694cd291dbc3d).
> According to the definition you propose it would have been a breaking
> change, wouldn't it? It just seems overly sensitive to say so.

Indeed, that would be a breaking change and I think that’s correct.  If
I use a stable version of nitrocli and `nitrocli otp set 1 name secret`
no longer works after a compatible update, I’d consider that a bug.

How would you define a breaking change?

> For that reason I actually like Cargo's model much more, which states
> > While SemVer says there is no compatibility before 1.0.0, Cargo
> > considers 0.x.y to be compatible with 0.x.z, where y ≥ z and x > 0.
> 
> That would at least have a meaning to users right now.

I didn’t add this to the PR because we have to define the public API
before we can think about compatibility guarantees.  But I agree that
this would be a good addition, especially in the Rust ecosystem.

d-e-s-o:
> Consider for instance the change of the default in the OTP format that we made a while back ([147d016](https://github.com/d-e-s-o/nitrocli/commit/147d01663f9b95c6819d3c6afe8694cd291dbc3d)). According to the definition you propose it would have been a breaking change, wouldn't it? It just seems overly sensitive to say so.
> Indeed, that would be a breaking change and I think that’s correct. If I use a stable version of nitrocli and `nitrocli otp set 1 name secret` no longer works after a compatible update, I’d consider that a bug. How would you define a breaking change?

I think what you propose is a reasonable definition. My main worry was that I didn't want to be forced into having to version bump for every small bit at this stage. Even more so given that I don't think that realistically anybody would care :D (but that's kind of an answer to the wrong question...). Anyway if we only care about such guarantees until after 1.0 I have no objection.

d-e-s-o:
I've merged the change into `devel`. Thanks!

PR #118, opened 2020-09-04
Introduce builder infrastructure for assembling Nitrocli instance

d-e-s-o:
In the future we would like to provide more ways for tests to create a
Nitrocli instance. In order to prevent explosion of with_XXX methods for
each possible combination of arguments, this change introduces a Builder
struct that can be used to create such an instance in an idiomatic way.

### DIFF
--- a/src/tests/config.rs
+++ b/src/tests/config.rs
@@ -34,17 +34,24 @@ $"#,
   )
   .unwrap();
 
-  let out = Nitrocli::with_model(model).handle(&["config", "get"])?;
+  let out = Nitrocli::make()
+    .model(model)
+    .build()
+    .handle(&["config", "get"])?;
+
   assert!(re.is_match(&out), out);
   Ok(())
 }
 
 #[test_device]
 fn set_wrong_usage(model: nitrokey::Model) {
-  let err = Nitrocli::with_model(model)
+  let err = Nitrocli::make()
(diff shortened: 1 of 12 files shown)


d-e-s-o:
Would you mind take a quick look please, Robin?

robinkrahl:
Overall, it looks good to me.  Some nitpicking that you don’t have to follow:  The `Nitrocli::make()` function that returns a `Builder` does not feel very idiomatic.  If we want to use a builder, I’d rather call `NitrocliBuilder::new()`.  But do we even need that?  Wouldn’t it be sufficient to add methods to `Nitrocli` that set the fields and return a mutable reference to self, i. e. use the [non-consuming builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html) like `std::process::Command`?

robinkrahl (approved):

d-e-s-o:
Thanks for taking a look! The issue with the non-consuming builder pattern in our case are the ergonomics: you can't just have a chain of configurations applied to an instance created in said chain and then assign the result, because Rust barks at you that a temporary is created and dropped. Anyway, I think I found a [reasonable compromise](https://github.com/d-e-s-o/nitrocli/commit/6bb521534727fadb58085acfca48433576293281) by making all "setters" consuming.

Issue #99, opened 2020-01-15
Dealing with multiple devices


robinkrahl:
As we now are able to detect multiple devices and select one of them to connect to, we should decide how to deal with multiple devices.

We already have the `--model` option and I think we should also have a `--serial-number` option, maybe even a `--usb-path` option, to select the Nitrokey device to connect to.

But what should we do if there are multiple devices but the user did not select one of them (or the user selected a model and there are multiple devices of that model)? Currently we just use the first one returned by hidapi.  This is the easiest solution, but it is also potentially dangerous.  My suggestion is to abort if we have more than one eligible device, asking the user to select one of them.  What do you think?

d-e-s-o:
> We already have the `--model` option and I think we should also have a `--serial-number` option, maybe even a `--usb-path` option, to select the Nitrokey device to connect to.

Agreed. They all seems useful and are orthogonal to each other.

> But what should we do if there are multiple devices but the user did not select one of them (or the user selected a model and there are multiple devices of that model)? Currently we just use the first one returned by hidapi. This is the easiest solution, but it is also potentially dangerous. My suggestion is to abort if we have more than one eligible device, asking the user to select one of them. What do you think?

Hm. It's a tough one. I haven't look at the changes to the API surface in detail (especially in `libnitrokey`), but it seems that this would imply having a new "list" call before connecting, correct? I am not a huge fan of doing so, mostly because it's an extra call (that's not a strong argument, but I'd say it is in line with how we kept things so far: close to the `libnitrokey` APIs, low overhead, and under the user's control; most recently with the `--no-connect` option to the `list` command, for example).

I'd also want to avoid context sensitive behavior where things work fine if nothing is specified but only a single device is plugged in, and then suddenly the program errors out if another stick is present. Such behavior tends to lead to unexpected issues at some point. But I guess one could argue that we already have that (you expect the status command to show information about the plugged in Pro stick but once you plugin in a Storage device that shows up by default, for example). To completely avoid such cases (I guess), we would require one of `{--model, --usb-path, --serial-number}` unconditionally. But that is swinging the pendulum too far away from the easy-to-use side, if you ask me.

My gut says that the vast majority of users wouldn't care one way or another, because they only have a single device (perhaps should be add telemetry reporting to get confirmation? :P). As such, I am tempted to say "let's just keep the existing behavior". But honestly, if you feel strongly about checking present devices upfront I am not opposed.

robinkrahl:
[...]

PR #119, opened 2020-09-06
Refactor connection handling for multiple devices

robinkrahl:
This patch series refactors the connection handling into two new functions, `find_device` and `connect`, and improves the handling of multiple attached devices as discussed in issue #99:  If mulitple attached devices match the filter set by the user, nitrocli now fails instead of just picking the first match.  And the `--serial-number` option now allows the user to filter the devices by serial number.  (We already have the --model option to filter by model.)

This means that we always list all attached Nitrokey devices before connecting.  Still the overhead is very small as the list can be read from the sys file system without any communication with the Nitrokey device itself.

Unfortunately, the Nitrokey Storage sends a wrong serial number in the USB device descriptor that is used for listing the devices, see https://github.com/Nitrokey/nitrokey-storage-firmware/issues/88.  Therefore the new `--serial-number` option only works for Nitrokey Pro devices.  Originally, I wanted to wait with this feature until this is fixed in the firmware, but I don’t think this will happen any time soon.  (And we would still have to support devices with older firmware versions.)

As a next step, we could add support for multiple devices to the `status` command.


d-e-s-o:
Thanks for the pull request! Looks good! I've made a few cosmetic changes and merged it. But can we come up with a way to test this functionality better? E.g., one way could be to have a test function that accepts any model (runs on pro & storage), lists the available devices to get their serial number, and then invokes `nitrocli` with `--serial-number` for each to see whether we connect to the correct device. Something similar could be done to check whether we bail out properly when multiple devices are attached. Do you think that's feasible?

robinkrahl:
Thanks!  I’ll check whether I can implement the test cases you
described.

However, your changes introduced a problem with the with_storage_device
function.  In my implementation, I always used Model::Storage when
calling connect.  Your implementation uses the model from the context
configuration, which may be empty if the user did not set the --model
option.  I’ll open a new PR with a fix.

d-e-s-o:
Oh, sorry about that :-| Didn't want too many review round trips. I'll merge your suggestions later today, but probably just as `fixup!` commits. Doesn't really matter to me in `devel`. Thanks for checking!

PR #120, opened 2020-09-07
Fix with_storage_device function

robinkrahl:
This PR fixes a bug in the connection logic of the with_storage_device function and makes the panic message used in the function more concise.


d-e-s-o:
Thanks for these fixes, Robin! I'll merge them into the respective commit to keep history somewhat clean in `master`.

PR #121, opened 2020-09-07
Add test cases for new connection logic

robinkrahl:
This patch adds two test cases for the new connection logic:
connect_multiple checks that nitrocli aborts if more than one Nitrokey
device is connected and no filter is set.  connect_serial_number checks
that we can use the --serial-number option to connect to all supported
Nitrokey devices.

---

This does not test the behaviour of the `--model` option if multiple
devices are connected, mainly because I only have one test device per
model.  If you have multiple devices, I can add a test case for that.

### DIFF
--- a/src/tests/run.rs
+++ b/src/tests/run.rs
@@ -3,6 +3,7 @@
 // Copyright (C) 2019-2020 The Nitrocli Developers
 // SPDX-License-Identifier: GPL-3.0-or-later
 
+use std::collections;
 use std::path;
 
 use super::*;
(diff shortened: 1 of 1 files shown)


robinkrahl:
I’ve added two more tests cases: for a wrong serial number and for the --model option, although I cannot verify the check for multiple devices of the same model.

d-e-s-o:
Thank you, Robin! Merged. Unfortunately I also can't test with multiple devices of the same model. The code looks good, though, so I guess we'll take the chance.

PR #122, opened 2020-09-08
Add --usb-path option to select device

robinkrahl:
This patch adds the --usb-path option as an additional way to filter the
Nitrokey device to connect to.  While the serial number is a better
identifier in theory, the Nitrokey Storage devices do not send their
serial number in the USB device descriptor.  Having the --usb-path
options allows users to select one of multiple Nitrokey Storage devices.

While we could directly call the nitrokey::Manager::connect_path
function with the specified path, we integrate the --usb-path option
into the existing find_device function for consistent error messages and
to avoid having to duplicate the --model and --serial-number checks.


src/config.rs
```diff
@@ -31,6 +31,8 @@ pub struct Config {
   #[merge(strategy = merge::vec::overwrite_empty)]
   #[serde(default, deserialize_with = "deserialize_serial_number_vec")]
   pub serial_numbers: Vec<nitrokey::SerialNumber>,
+  /// The usb path of the device to connect to.
```
d-e-s-o, line 34: *USB

src/commands.rs
```diff
@@ -75,16 +78,17 @@ fn find_device(config: &config::Config) -> anyhow::Result<nitrokey::DeviceInfo>
...
 
   let device = iter
     .next()
     .with_context(|| format!("Nitrokey device not found{}", format_filter(config)))?;
 
   anyhow::ensure!(
     iter.next().is_none(),
-    "Multiple Nitrokey devices found{}.  Use the --model and --serial-number options to \
-    select one",
+    "Multiple Nitrokey devices found{}.  Use the --model, --serial-number and --usb-path options \
```
d-e-s-o, line 90: I think in American English (which I try to use here) we'd use a comma before 'and' or 'or' in enumerations with more than two elements (https://en.wikipedia.org/w/index.php?title=Serial_comma&oldid=976699191).

src/tests/run.rs
```diff
@@ -142,6 +143,26 @@ fn connect_wrong_serial_number(_model: nitrokey::Model) {
   );
 }
 
+#[test_device]
+fn connect_usb_path(_model: nitrokey::Model) -> anyhow::Result<()> {
+  let devices = nitrokey::list_devices()?;
+  for path in devices.iter().map(|d| &d.path) {
+    let res = Nitrocli::new().handle(&["status", &format!("--usb-path={}", path)]);
+    assert!(res.is_ok());
```
d-e-s-o, line 151: Could we do the same you did with the serial number where you check the serial number in the status output against the expected one taken from the device we connect to?

d-e-s-o (commented):
Thanks for the pull request, Robin! I have a few minor comments/questions.

doc/nitrocli.1
```diff
@@ -12,11 +12,13 @@ It can be used to access the encrypted volume, the one-time password generator,
...
 .SS Device selection
 Per default, \fBnitrocli\fR connects to any attached Nitrokey device.
-You can use the \fB\-\-model\fR and \fB\-\-serial-number\fR options to select
-the device to connect to.
+You can use the \fB\-\-model\fR, \fB\-\-serial-number\fR and \fB\-\-usb-path\fR
+options to select the device to connect to.
 \fBnitrocli\fR fails if more than one attached Nitrokey device matches this
 filter or if multiple Nitrokey devices are attached and none of the filter
 options is set.
+Use the \fBlist\fR command to list all attached devices with their USB path,
```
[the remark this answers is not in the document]
robinkrahl: Yes!  This reminds me that I forgot to address the naming of the option in the PR description:

The path value originates in hidapi as the `path` field of the `hid_device_info` struct and is described as the “Platform-specific device path”.  In libnitrokey, it’s the `path` field of the `NK_device_info` struct.  In nitrokey-rs, it’s the `path` field of the `DeviceInfo` struct.

I did not want to use `--path` as the option name because it is very generic and I would expect a file system path as its value.  `--device-path` could have been an option, but then I would expect something like `/dev/hidraw0`.  At least on my machine, the device path is a combination of the USB bus, the USB device and (probably) the HID device index, e. g. `004:001:00`.  Therefore I decided to use the term USB path although, as far as I know, no-one else is using it.  What do you think?

src/tests/run.rs
robinkrahl: The status currently does not include the device path because we use the `with_device` function that does not expose the `DeviceInfo` with the path.  I’m currently preparing a patch that makes the `status` command print the status for all attached devices.  We will have to call `list` directly, so we will be able to include the device path in the output.  But for now, I think it is the easiest solution to just make sure that the command is successful.

robinkrahl:
Thanks for the review!  I added fixes for all issues.

doc/nitrocli.1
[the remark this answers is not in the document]
d-e-s-o: I'd say it's fine. It's explained pretty well in the man page and I share your concerns about "device path" being ambiguous, too (and probably more so). So as long as it's consistent within the program I think it's okay.

src/tests/run.rs
d-e-s-o: What I mean is (roughly):
```rust
  for path in devices.iter().map(|d| &d.path) {
    let serial_number = d.serial_number.unwrap();
    let output = Nitrocli::new().handle(&["status", &format!("--usb-path={}", path)])?;
    assert!(res.contains(&format!("serial number:     {}\n", sn)));
```

So instead of verifying the USB path directly just take the route over the serial number. Still won't provide better coverage for storage devices, but I think it's fine. Once (or perhaps: if) storage devices report their serial number properly it will just work.

robinkrahl: Okay, done.

d-e-s-o:
> Thanks for the review! I added fixes for all issues.

I'll merge it later today. Thanks!

d-e-s-o:
All in now! Thanks again Robin!

PR #123, opened 2020-09-09
Support mulitple devices in status command

robinkrahl:


d-e-s-o:
Sorry Robin, I am running a bit behind on this one. Will take a look later tomorrow.

src/commands.rs
```diff
@@ -65,11 +65,11 @@ fn format_filter(config: &config::Config) -> String {
   }
 }
 
-/// Find a Nitrokey device that matches the given requirements
-fn find_device(config: &config::Config) -> anyhow::Result<nitrokey::DeviceInfo> {
+/// Find all Nitrokey devices that match the given requirements
+fn find_devices(config: &config::Config) -> anyhow::Result<Vec<nitrokey::DeviceInfo>> {
```
d-e-s-o, line 69: Hm. I'd not want to introduce unnecessary allocations where they can be avoided (and yes, I do realize `nitrokey-rs` already allocates a vector). We could remove the need for a `Vec` here with the following patch:
```diff
--- src/commands.rs
+++ src/commands.rs
@@ -65,23 +65,27 @@ fn format_filter(config: &config::Config) -> String {
   }
 }

-/// Find all Nitrokey devices that match the given requirements
-fn find_devices(config: &config::Config) -> anyhow::Result<Vec<nitrokey::DeviceInfo>> {
-  let devices = nitrokey::list_devices().context("Failed to enumerate Nitrokey devices")?;
+/// Filter out all Nitrokey devices that match the given requirements
+fn filter_devices<'c, I>(
+  devices: I,
+  config: &'c config::Config,
+) -> impl Iterator<Item = nitrokey::DeviceInfo> + 'c
+where
+  I: ExactSizeIterator<Item = nitrokey::DeviceInfo> + 'c,
+{
   let nkmodel = config.model.map(nitrokey::Model::from);
-  let matching_devices = devices
-    .into_iter()
-    .filter(|device| nkmodel.is_none() || device.model == nkmodel)
-    .filter(|device| {
+  devices
+    .filter(move |device| nkmodel.is_none() || device.model == nkmodel)
+    .filter(move |device| {
       config.serial_numbers.is_empty()
         || device
           .serial_number
           .map(|sn| config.serial_numbers.contains(&sn))
           .unwrap_or_default()
     })
-    .filter(|device| config.usb_path.is_none() || config.usb_path.as_ref() == Some(&device.path))
-    .collect();
-  Ok(matching_devices)
+    .filter(move |device| {
+      config.usb_path.is_none() || config.usb_path.as_ref() == Some(&device.path)
+    })
 }

 /// Connect to a Nitrokey device that matches the given requirements
@@ -89,18 +93,20 @@ fn connect<'mgr>(
   manager: &'mgr mut nitrokey::Manager,
   config: &config::Config,
 ) -> anyhow::Result<nitrokey::DeviceWrapper<'mgr>> {
-  let device_infos = find_devices(config)?;
+  let devices = nitrokey::list_devices().context("Failed to enumerate Nitrokey devices")?;
+  let mut device_infos = filter_devices(devices.into_iter(), config);
+
+  let device = device_infos
+    .next()
+    .with_context(|| format!("Nitrokey device not found{}", format_filter(config)))?;

   anyhow::ensure!(
-    device_infos.len() < 2,
+    device_infos.next().is_none(),
     "Multiple Nitrokey devices found{}.  Use the --model, --serial-number, and --usb-path options \
     to select one",
     format_filter(config)
   );

-  let device = device_infos
-    .first()
-    .with_context(|| format!("Nitrokey device not found{}", format_filter(config)))?;
   manager.connect_path(device.path.deref()).with_context(|| {
     format!(
       "Failed to connect to Nitrokey device at path {}",
```

What do you think?

doc/nitrocli.1
```diff
@@ -67,10 +67,12 @@ To omit the serial number of Nitrokey Storage devices instead of connecting to
...
 .TP
 .B nitrocli status
-Print the status of the connected Nitrokey device, including the stick serial
-number, the firmware version, and the PIN retry count. If the device is a
-Nitrokey Storage, also print storage related information including the SD card
-serial number, the encryption status, and the status of the volumes.
+Print the status of the attached Nitrokey devices that match the filter
+options (see the Device selection section), including the USB path, stick
+serial number, the firmware version, and the PIN retry count.
+If the device is a Nitrokey Storage, also print storage related information
```
d-e-s-o, line 73: Perhaps rather "If **a** device is [...]"?

src/commands.rs
```diff
@@ -363,19 +363,27 @@ fn print_storage_status(
...
+) -> anyhow::Result<()> {
   let serial_number = device
     .get_serial_number()
     .context("Could not query the serial number")?;
 
   println!(
     ctx,
-    r#"Status:
+    r#"Device #{idx}:
+  USB path:          {path}
```
d-e-s-o, line 379: Would you mind creating a separate change for the printing of the path? Or does that not make sense?

src/commands.rs
```diff
@@ -402,7 +410,22 @@ fn print_status(ctx: &mut Context<'_>, device: &nitrokey::DeviceWrapper<'_>) ->
 
 /// Inquire the status of the nitrokey.
 pub fn status(ctx: &mut Context<'_>) -> anyhow::Result<()> {
-  with_device(ctx, |ctx, device| print_status(ctx, &device))
+  let device_infos = find_devices(&ctx.config)?;
```
d-e-s-o, line 413: Is the order in which devices are returned by `nitrokey-rs` guaranteed to be stable between invocations?

src/commands.rs
```diff
@@ -402,7 +410,22 @@ fn print_status(ctx: &mut Context<'_>, device: &nitrokey::DeviceWrapper<'_>) ->
...
+    "Nitrokey device not found{}",
+    format_filter(&ctx.config)
+  );
+
+  let mut manager =
+    nitrokey::take().context("Failed to acquire access to Nitrokey device manager")?;
+  for (idx, device_info) in device_infos.iter().enumerate() {
+    let device = manager
+      .connect_path(device_info.path.deref())
+      .context("Failed to connect to Nitrokey")?;
```
d-e-s-o, line 425: Would it make sense to include the path in the error here?

src/tests/status.rs
```diff
@@ -64,3 +66,19 @@ $"#,
   assert!(re.is_match(&out), out);
   Ok(())
 }
+
+#[test_device]
+fn output_multiple(_model: nitrokey::Model) -> anyhow::Result<()> {
+  let devices = nitrokey::list_devices()?;
+  let res = Nitrocli::new().handle(&["status"])?;
```
d-e-s-o, line 73: I guess I missed that before, but can we use `out` or some other name instead of `res`? The only association I have with `res` is result and in Rust code I typically try to use that only for `Result` objects. In any case, I believe something indicating that we are dealing with the program output makes more sense here.

d-e-s-o (commented):
Thanks for the pull request, Robin! Do you intend to adjust other commands to also apply to multiple sticks plugged in in the future? If not, why is `status` special? I am a little worried that what we are doing here is causing more harm than good. Is there more background to this pull request? So far I found the model that `nitrocli` follows very easy for users to understand: we work on a single device. Always. Now it gets a bit wishy-washy, it feels like.

robinkrahl:
No, I don’t intend to change other commands.

> We work on a single device. Always.

That’s not true:  The `list` command already works on multiple devices.  I think the `status` command is similar to `list`.  It is  typically used to get an overview of the environment.  I would consider it a bad UI if I had to first call `nitrocli list` and then `nitrocli status` for every device.

At the same time, I don’t see any harm this could cause.  Sure, listing several devices may take a bit longer, but all operations are read-only and don’t require authentication.

src/commands.rs
robinkrahl: I thought about this, but I decided that I’d be willing to pay for easier code with one vector allocation.  But I can change that if you want to.

src/commands.rs
robinkrahl: Yes, I can do that.

src/commands.rs
robinkrahl: Not by nitrokey-rs.  The order is determined by libnitrokey, which orders by USB vendor ID, then by the order returned by hidapi’s `hid_enumerate` function.  hidapi’s documentation does not mention the order.

src/commands.rs
robinkrahl: Sure.

src/tests/status.rs
robinkrahl: Sure.

d-e-s-o:
So why do we need `list` at all then? Isn't it just reporting a subset of what `status` is showing now?

robinkrahl:
> So why do we need `list` at all then? Isn't it just reporting a subset
> of what `status` is showing now?

Yes, but:
1. `list` does not have to connect to the Nitrokey devices, so it is
   much faster.  This is especially useful for extensions that want to
   perform an operation on all connected devices (e. g. check the
   available OTP slots).
2. `list` shows exactly the same information that is used by the
   `--model`, `--usb-path` and `--serial-number` options.

d-e-s-o:
It still seems like a confusing duplication to me. And in my opinion that's not great UI either. Would it make sense to remove the `status` command and merge everything into `list` and then add a flag that switches between the two formats? Presumably, the more verbose version would be the default, as it is somewhat more useful, but I can see why extensions may want to use the concise output.

d-e-s-o:
Seems the discussion has stalled. Closing.

PR #126, opened 2020-09-19
Add the --progress option to the fill command

robinkrahl:
The fill command starts a background operation on a Nitrokey Storage
device that fills the SD card with random data.  This patch adds a new
option, --progress, to the fill command that checks if a fill operation
is already running on the device and shows its progress.


d-e-s-o:
I'll look at this pull request tomorrow.

d-e-s-o (commented):
Great change! Thanks for adding this functionality. What do you think about renaming the option to `--resume` (or `--attach` that you brought up elsewhere)?
To all the non-manual and non-help reading users I think "progress" intuitively conveys a slightly different meaning: that of showing progress, which is something we do by default (where supported).

robinkrahl:
> What do you think about renaming the option to `--resume` (or
> `--attach` that you brought up elsewhere)?

I think --resume is problematic because it implies that the operation is
currently paused.  I didn’t use --attach because I feared that it might
be unclear for users without reading the documentation, but better
unclear than misleading. ;)  I’ll push a fix soon.

d-e-s-o:
Merged! Thanks Robin!

PR #138, opened 2021-03-26
Use new PWS access API

### LINKED ISSUE #133: Retrieving password fails when "Login name" field is left empty
I have a password for which I do not have a "login" and nitrocli fails to retrieve the values of that slot.

`nitrocli pws get 1` results in:
```
name:     nologinpw
Command error: The given slot is not programmed
```

Device: Nitrokey Pro 2

robinkrahl:
In previous versions of nitrokey-rs, the getters for the name, login and
password fields of a PWS slot would return an error for empty strings as
libnitrokey uses empty strings to indicate unprogrammed slots.  But as
the user might have deliberately specified an empty value, nitrokey-rs
v0.9.0 introduces a new PWS API that makes it possible to distinguish
unprogrammed slots and empty values.

This patch updates the nitrokey-rs dependency and refactors the pws
status and pws get commands to use the new PWS API.

Fixes #133.

To discuss:
- Do we want to support writing completely empty PWS slots, i. e. no name, login or password?

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -4,7 +4,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 use std::borrow;
-use std::convert::TryFrom as _;
 use std::env;
 use std::ffi;
 use std::fmt;
@@ -996,20 +995,6 @@ fn print_pws_data(
   Ok(())
 }
 
-fn check_slot(pws: &nitrokey::PasswordSafe<'_, '_>, slot: u8) -> anyhow::Result<()> {
-  if slot >= nitrokey::SLOT_COUNT {
-    anyhow::bail!("Slot {} is not valid", slot);
-  }
-  let status = pws
-    .get_slot_status()
-    .context("Failed to read PWS slot status")?;
-  if status[slot as usize] {
-    Ok(())
-  } else {
-    anyhow::bail!("Slot {} is not programmed", slot)
-  }
-}
-
 /// Read a PWS slot.
 pub fn pws_get(
   ctx: &mut Context<'_>,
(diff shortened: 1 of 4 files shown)


robinkrahl:
@d-e-s-o Could you please review the changes?  If you don’t have any objections, I’ll release a new nitrokey-rs version soon and then convert this draft into a PR.

d-e-s-o:
Thanks for the pull request, Robin! I'll take a closer look hopefully tomorrow morning and will hopefully have an answer to your question by then.

src/commands.rs
```diff
@@ -1020,17 +1005,17 @@ pub fn pws_get(
   quiet: bool,
 ) -> anyhow::Result<()> {
   with_password_safe(ctx, |ctx, pws| {
-    check_slot(&pws, slot).context("Failed to access PWS slot")?;
+    let slot = pws.get_slot(slot).context("Failed to access PWS slot")?;
```
d-e-s-o, line 1008: Can you please check that we still report a proper user-understandable error in all three cases? I think we should be good, but doesn't hurt to double check.

src/commands.rs
```diff
@@ -1060,31 +1045,29 @@ pub fn pws_clear(ctx: &mut Context<'_>, slot: u8) -> anyhow::Result<()> {
...
   Ok(())
 }
 
 /// Print the status of all PWS slots.
 pub fn pws_status(ctx: &mut Context<'_>, all: bool) -> anyhow::Result<()> {
   with_password_safe(ctx, |ctx, pws| {
-    let slots = pws
-      .get_slot_status()
-      .context("Failed to read PWS slot status")?;
+    let slots = pws.get_slots().context("Failed to read PWS slot status")?;
```
d-e-s-o, line 1063: Consider adjusting the error message slightly to `"Failed to read PWS slots"`.

d-e-s-o (commented):
Looks great to me and like the right step. I'd think we can just support setting everything empty like you added a test case for. As long as nothing bad will happen, the UI seems reasonably clear to me.

src/commands.rs
robinkrahl: We now get the nitrokey-rs error variants:

```
$ nitrocli pws get 20
Library error: The given slot is invalid
$ nitrocli pws get 15
Command error: The given slot is not programmed
```

I think that’s fine.

src/commands.rs
robinkrahl: I’m not sure about that.  Internally, we still call the `GET_PW_SAFE_SLOT_STATUS` command.  “Failed to read PWS slots” sounds as if we tried to read the slot values.

robinkrahl:
Thanks for the review!  I’ve released nitrokey-rs v0.9.0 and updated the PR.

src/commands.rs
robinkrahl: Though we lose the context when [downcasting](https://github.com/d-e-s-o/nitrocli/blob/211a25421c879097ec3a8b5e030795d5f84c7261/src/commands.rs#L277) the error.  As `downcast` consumes the original error, I’m not sure how to fix this.

d-e-s-o:
I merged the change into `devel`. Thanks for fixing up this issue!

PR #142, opened 2021-04-09
Allow spaces in base32 OTP secrets

### LINKED ISSUE #136: Allow spaces in OTP base32 secret
Some sites display the base32 secret for OTP authentication in groups, separated by spaces.  nitrocli should be able to handle such input without requiring the user to manually remove the spaces.

robinkrahl:
Some sites as GitLab display their OTP secrets as base32 strings in
groups separated by spaces.  Previously, the user had to manually remove
the spaces before passing the secret to nitrocli.  But as the space can
never be part of the base32 string itself, we can unambigously identify
it as a separator and remove it from the secret before processing it.

Fixes #136.


d-e-s-o:
Merged as well! Thanks for identifying and working on this issue.

PR #143, opened 2021-04-10
Add .editorconfig file for Rust indentation

robinkrahl:
While most editors use the standard 4-space indentation for Rust code,
this project uses two spaces.  To make it easier to write code according
to the project style, this patch adds an .editorconfig file that
specifies the indentation rules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# .editorconfig
+
+# Copyright (C) 2021 The Nitrocli Developers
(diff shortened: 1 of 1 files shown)


d-e-s-o:
Merged as well. Thanks, was not aware of editor config!

robinkrahl:
> Merged as well.

Thanks!

> Thanks, was not aware of editor config!

Feels much less invasive and risky than `set exrc`. :-)

PR #145, opened 2021-04-11
Fix cache ID for pinentry calls

robinkrahl:
Since nitrokey-rs v0.8.0 added support for the Librem Key, the Display
implementation for nitrokey::Model may include a space, for example
"Nitrokey Pro" (instead of "Pro").  I missed that we used this string
representation to generate the pinentry cache ID, leading to an
additional space in the GnuPG GET_PASSPHRASE call.  This messed up the
error message, prompt and description arguments that come after the
cache ID.

With this patch, we use a hard-coded string for the cache IDs instead of
nitrokey::Model’s Display implementation.

### DIFF
--- a/src/pinentry.rs
+++ b/src/pinentry.rs
@@ -58,7 +58,12 @@ impl PinEntry {
 
 impl SecretEntry for PinEntry {
   fn cache_id(&self) -> Option<CowStr> {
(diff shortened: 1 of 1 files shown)


d-e-s-o:
Fabulous stuff! I've seen that issue a bunch of times myself, but had not investigated yet. Merged.

PR #146, opened 2021-04-12
Add otp-cache extension

robinkrahl:
This PR is an iteration of PR #144: Add cache extension.  It introduces the
`NITROCLI_RESOLVED_USB_PATH` environment variable to avoid having to duplicate
the device selection logic, it moves the `ext.rs` module into a new extension
support crate, `nitrokey-ext`, and adds the nitrocli-otp-cache extension.

### DIFF
--- a/ext/ext/src/lib.rs
+++ b/ext/ext/src/lib.rs
@@ -1,0 +1,141 @@
+// lib.rs
+
+// Copyright (C) 2020-2021 The Nitrocli Developers
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+use std::env;
+use std::ffi;
+use std::process;
+
+use anyhow::Context as _;
+
+#[derive(Debug)]
+pub struct Context {
+  pub nitrocli: ffi::OsString,
+  pub resolved_usb_path: Option<String>,
+  pub verbosity: Option<u8>,
+  pub project_dirs: directories::ProjectDirs,
+}
+
+impl Context {
+  pub fn from_env(name: &str) -> anyhow::Result<Self> {
+    let nitrocli = env::var_os("NITROCLI_BINARY")
+      .context("NITROCLI_BINARY environment variable not present")
+      .context("Failed to retrieve nitrocli path")?;
+
+    let resolved_usb_path = env::var("NITROCLI_RESOLVED_USB_PATH").ok();
+
+    let verbosity = env::var_os("NITROCLI_VERBOSITY")
+      .context("NITROCLI_VERBOSITY environment variable not present")
+      .context("Failed to retrieve nitrocli verbosity")?;
+    let verbosity = if verbosity.len() == 0 {
+      None
+    } else {
+      let verbosity = verbosity
+        .to_str()
+        .context("Provided verbosity string is not valid UTF-8")?;
+      let verbosity = u8::from_str_radix(verbosity, 10).context("Failed to parse verbosity")?;
+      set_log_level(verbosity);
+      Some(verbosity)
+    };
+
+    let project_dirs = directories::ProjectDirs::from("", "", name).with_context(|| {
+      format!(
+        "Could not determine the application directories for the {} extension",
+        name
+      )
+    })?;
+
+    Ok(Self {
+      nitrocli,
+      resolved_usb_path,
+      verbosity,
+      project_dirs,
+    })
+  }
+
+  pub fn nitrocli(&self) -> Nitrocli {
+    Nitrocli::from_context(self)
+  }
+
+  pub fn connect<'mgr>(
+    &self,
+    mgr: &'mgr mut nitrokey::Manager,
+  ) -> anyhow::Result<nitrokey::DeviceWrapper<'mgr>> {
+    if let Some(usb_path) = &self.resolved_usb_path {
+      mgr.connect_path(usb_path.to_owned()).map_err(From::from)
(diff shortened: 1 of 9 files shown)


ext/ext/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+use std::process;
+
+use anyhow::Context as _;
+
+#[derive(Debug)]
+pub struct Context {
+  pub nitrocli: ffi::OsString,
+  pub resolved_usb_path: Option<String>,
+  pub verbosity: Option<u8>,
+  pub project_dirs: directories::ProjectDirs,
```
d-e-s-o, line 17: What do you think about making these members private? Feels like they are implementation details for the most part. The project directories I'd consider exposing just as paths, so that we don't expose some third party crate's data type across crate boundaries: `Context::cache_dir() -> &Path` (just expose what we need right now).

What do you think?

ext/ext/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+impl Context {
+  pub fn from_env(name: &str) -> anyhow::Result<Self> {
```
d-e-s-o, line 21: Would you mind adding a method header? The `name` may not be self explanatory without looking at the implementation, I'd say.

ext/otp-cache/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+struct Slot {
+  name: String,
+  id: u8,
+}
+
+/// Access Nitrokey OTP slots by name
+#[derive(Debug, structopt::StructOpt)]
+#[structopt(bin_name = "nitrocli cache")]
```
d-e-s-o, line 26: `otp-cache`

ext/otp-cache/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+fn cmd_get(
+  ctx: &nitrocli_ext::Context,
+  cache_file: &path::Path,
+  slot_name: &str,
+) -> anyhow::Result<()> {
+  let cache = get_cache(cache_file)?;
+  let totp_slots: Vec<_> = cache.totp.iter().filter(|s| s.name == slot_name).collect();
+  let hotp_slots: Vec<_> = cache.hotp.iter().filter(|s| s.name == slot_name).collect();
+  if totp_slots.len() + hotp_slots.len() > 1 {
+    Err(anyhow::anyhow!("Multiple OTP slots with the given name"))
```
d-e-s-o, line 80: Nit: `"Found multiple OTP slots with the given name"`?

ext/otp-cache/src/main.rs
```diff
@@ -0,0 +1,174 @@
...
+  } else if let Some(slot) = totp_slots.first() {
+    generate_otp(&ctx, "totp", slot.id)
+  } else if let Some(slot) = hotp_slots.first() {
+    generate_otp(&ctx, "hotp", slot.id)
+  } else {
+    Err(anyhow::anyhow!("No OTP slot with the given name"))
```
d-e-s-o, line 86: Similar comment here, add verb?

d-e-s-o (commented):
I think it looks great! Thanks for splitting and reworking the earlier pull request! I may only get to take a final look and be able to merge it later this week (around Thursday or later).

ext/ext/src/lib.rs
robinkrahl: You mean a doc comment?  Sure, we probably should do that for the entire module.  I also wasn’t sure if we should automatically deduce the name from `env::args()` or `env::current_exe()`, or at least provide a method that does that.  What do you think?

robinkrahl:
Thanks for the review!  I’ll fix the issues and update the PR later today.

ext/ext/src/lib.rs
d-e-s-o: Yeah. Automatic deduction seems fine to me. Let's do just that for now, as it's all we need.

robinkrahl:
I think I’ve addressed all issues.

d-e-s-o:
Hi Robin, why do you think we should have a dedicated `update` command? Conceptually I am fine with it (just as an alternative to `--force-update`), but it seems unnecessary to require users to use it. Can you please share your thoughts?

robinkrahl:
Good point!  The cache lookup logic became pretty convoluted so I decided to separate reading and writing the cache.  (For example we have to make sure that the `nitrokey::Manager` is dropped before calling `nitrocli`.)  But I‘m fine with including the update logic in the get command too.  I’m not sure which way is more intuitive and useful.

d-e-s-o:
Sounds good. I vote for integration.

> "There is no cached slot data. Run the update command to initialize the cache."

If a computer can tell me precisely what to do most of the time I just want it to do it for me. It was definitely the case for me here.

robinkrahl:
I’ve removed the `update` command.

robinkrahl:
Do we want separate man pages for extensions?

d-e-s-o:
Awesome, thank you!

> Do we want separate man pages for extensions?

It's definitely not a priority right now for me. I'd hope we could cover most of it with a some decent help text. I'd rather we think about tests first on the extension front.

d-e-s-o:
Fantastic work, Robin! Merged into `devel`.

PR #148, opened 2021-04-13
Add libnitrokey version to -V/--version output

### LINKED ISSUE #147: List libnitrokey version
Knowing the libnitrokey version used by nitrocli is an important information for debugging.  Currently, it is impossible to determine the library version unless nitrocli is linked against a shard libnitrokey.  We should probably have a version command that lists both the nitrocli version and the libnitrokey version as returned by `nitrokey::get_library_version()`.

robinkrahl:
Currently, it is hard to determine the libnitrokey version used by
nitrocli unless it is linked against a shared library.  This patch adds
the libnitrokey version (nitrokey::get_library_version()) to the
-V/--version output.

Fixes #147.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -8,7 +8,7 @@ use std::ffi;
 
 /// Provides access to a Nitrokey device
 #[derive(Debug, structopt::StructOpt)]
-#[structopt(name = "nitrocli")]
+#[structopt(name = "nitrocli", no_version)]
 pub struct Args {
   /// Increases the log level (can be supplied multiple times)
   #[structopt(short, long, global = true, parse(from_occurrences))]
--- a/src/main.rs
+++ b/src/main.rs
@@ -110,8 +110,11 @@ impl error::Error for DirectExitError {}
 
 /// Parse the command-line arguments and execute the selected command.
 fn handle_arguments(ctx: &mut Context<'_>, argv: Vec<String>) -> anyhow::Result<()> {
-  match args::Args::from_iter_safe(argv.iter()) {
-    Ok(args) => {
+  let version = get_version_string();
+  let clap = args::Args::clap().version(version.as_str());
+  match clap.get_matches_from_safe(argv.iter()) {
+    Ok(matches) => {
+      let args = args::Args::from_clap(&matches);
       ctx.config.update(&args);
       args.cmd.execute(ctx)
     }
@@ -154,6 +157,15 @@ fn handle_arguments(ctx: &mut Context<'_>, argv: Vec<String>) -> anyhow::Result<
         Ok(())
       }
     }
+  }
+}
+
+fn get_version_string() -> String {
+  let version = env!("CARGO_PKG_VERSION");
+  if let Ok(library_version) = nitrokey::get_library_version() {
+    format!("{} using libnitrokey {}", version, library_version)
+  } else {
+    format!("{} using an undetectable libnitrokey version", version)
   }
 }
 
--- a/src/tests/run.rs
+++ b/src/tests/run.rs
@@ -100,7 +100,7 @@ fn version_option() {
     assert!(re.is_match(&s), out);
   }
 
-  let re = regex::Regex::new(r"^nitrocli \d+.\d+.\d+(-[^-]+)*\n$").unwrap();
+  let re = regex::Regex::new(r"^nitrocli \d+.\d+.\d+(-[^-]+)* using libnitrokey .*\n$").unwrap();
 
   test(&re, "--version");
   test(&re, "-V");


robinkrahl:
Apparently, the CI is no longer running for PRs since last September.  Is that intentional?

d-e-s-o:
> Apparently, the CI is no longer running for PRs since last September. Is that intentional?

It's intentional in the sense that I destroyed the AWS VM hosting the necessary infrastructure. If it were still working or suddenly worked again I'd be scratching my head a bit. That being said, I looked into getting it back up again after the uptick in pull requests -- self hosted for a change. It's not clear whether it will work, but we shall see. Earliest by the weekend, I suppose.

d-e-s-o:
Merged. Thanks Robin!

robinkrahl:
Could you please remind me why you are using Gitlab for CI? If you would consider using a different service I could assist configuring it.

d-e-s-o:
> Could you please remind me why you are using Gitlab for CI? If you would consider using a different service I could assist configuring it.

I decided for GitLab because it was the only one that fit the bill. See 6d0db4ef0d3c1cde9cc98606cbbb32977ece4218. Aside from the reasons mentioned in there, most others didn't even allow configuration of the configuration file (which at least back then was a necessity). Back when I went with GitLab it was unfortunately not obvious that pull requests were not supported and it's still non-obvious (presumably intentionally so...).

I think requirements as well as products have changed. GitHub Actions (non-existent back in the day) may fit the bill (haven't checked) and we could consider switching to it if it works for everything. I will probably not switch to some higgledy-piggledy CI that I (subjectively) have to worry may die any day or cease to have a reasonable free tier. I am not particularly attached to GitLab, other than using it for all my projects right now. I am not opposed to a one-off switch (of `nitrocli`) if it provides value.

robinkrahl:
I don’t really care where the CI is running.  Using a self-hosted runner
just sounds like even more maintenance effort and even more potential
for errors or incompatibilities.  So if you want to I can try to
prepare a GitHub Actions configuration.  If you want to stick with
Gitlab that’s okay too.

d-e-s-o:
I wouldn't be hosting my own runner. If you have the cycles feel free to give GitHub Actions a try. I'd certainly be happy having one less platform we depend on.

robinkrahl:
I’ve made a first attempt [here](https://github.com/robinkrahl/nitrocli/commit/a0ffc5b5c07c6c9fa929a1dec16758826a8e2aa6) ([logs](https://github.com/robinkrahl/nitrocli/runs/2354548431)).  I think it’s working pretty well:  Of course, `apt-get` does not work on MacOS and Windows, so I’d drop these platforms until someone volunteers to write the tests.  The tests fail on Rust nightly, but this seems to be a legitimate issue.  And clippy has some warnings.

My only issue is that GitHub does not show the full build logs unless you are logged in.  But still better to have restricted access to the build logs than having no builds at all. ;-)

d-e-s-o:
Very cool! Thanks for working on that.

> Of course, apt-get does not work on MacOS and Windows, so I’d drop these platforms until someone volunteers to write the tests.

Sounds good.

Can we consider using `https://github.com/actions-rs/toolchain` ([larger article](https://dev.to/wayofthepie/rust-and-github-actions-3kob)) instead of `hecrj`'s stuff, though? Seems a bit more legit. Do you see value in testing on `stable, beta, nightly`? I think all we care about is the minimum supported Rust version. If we use the latest `stable` we'll probably quickly get into a situation where the more recent `clippy` complains about something that we'd need to fix that in turn relies of a newer version of Rust (e.g., `match!`). (though I suppose we could pick an older `clippy` with a newer toolchain, but the value proposition is not super clear to me, off-hand)

Shame they don't support Debian...

Do you know the reasoning behind `fail-fast: false`? Or is that mostly a remnant from copy & pasting?

robinkrahl:
> Can we consider using `https://github.com/actions-rs/toolchain`

Sure!  I’ll have a look at it.  The action infrastructure still fills a
bit strange to me.  I’d rather use system packages to install software,
but unfortunately, rustup is not packaged for Ubuntu.

> Do you see value in testing on `stable, beta, nightly`?

I think we should definitely test stable.  And it’s good to catch issues
as early as possible, e. g. this one:
	https://github.com/robinkrahl/nitrocli/runs/2354548795#step:7:446
So I would rather ask:  Why not?

> If we use the latest `stable` we'll probably quickly get into a
> situation where the more recent `clippy` complains about something
> that we'd need to fix that in turn relies of a newer version of Rust

Currently, only build and test are run for all Rust versions.  clippy is
always running on stable.  We could replace that with the MSRV to avoid
compatibility issues.

> Shame they don't support Debian...

True.  Or at least any other distribution like Arch.

> Do you know the reasoning behind `fail-fast: false`?

I find it useful to know in which Rust versions an issue occurs.  For
example, it is probably useful to know that an issue only occurs on
nightly but not on stable.  So I’d like all runs to finish even if one
fails.

d-e-s-o:
> Do you see value in testing on `stable, beta, nightly`?
> I think we should definitely test stable. And it’s good to catch issues as early as possible, e. g. this one: https://github.com/robinkrahl/nitrocli/runs/2354548795#step:7:446 So I would rather ask: Why not?

For the sake of the planet? :D And...Microsoft's energy bill. At least with beta and nightly I don't see value but potential for noise/spurious failures.

> Do you know the reasoning behind `fail-fast: false`?
> I find it useful to know in which Rust versions an issue occurs. For example, it is probably useful to know that an issue only occurs on nightly but not on stable. So I’d like all runs to finish even if one fails.

:+1:

robinkrahl:
I think once we enable caching, the overhead for the additional
toolchains should be rather low.  But I don’t insist on testing nightly
and beta.

> And...Microsoft's energy bill.

Don’t tempt me!

PR #149, opened 2021-04-14
Add --only-aes-key option to reset command

### LINKED ISSUE #69: Support factory reset
I’m unsure how to name the factory reset command:  Based on the existing commands, `reset` is a fitting choice.  On the other hand, it might be beneficial to have a name that is more explicit, like `factory-reset`.  Given that we also have to name the command to build the AES key, we probably have to start using more complex command names anyway.  What do you think?

robinkrahl:
This patch adds an --only-aes-key option to the reset command to only
build a new AES key without performing a full factory reset.

Fixes #69.

----

Unfortunately, I can’t get the new unit test to work.  I get a Wrong Password error for the reset call.  But if I manually execute the command, it’s working fine.  Can you confirm this problem?  Do you see the reason for the error?


robinkrahl:
Nevermind, applying the same workaround as for the regular password reset helped.

d-e-s-o:
Thanks for the pull request, Robin! I'll take a closer look at this and your other work later this week. Running a bit thing right now.

d-e-s-o:
Great work! Merged this one, too, but I am a bit confused by your usage of "admin PIN" in the test. I think you meant user PIN and changed it, but perhaps I am missing something.

robinkrahl:
You’re totally right.  I started out using the admin PIN before I realized that we need the user PIN to enable the PWS and then forgot to fix the comment.  Thanks!

Issue #139, opened 2021-04-04
Improve pws subcommand


robinkrahl:
Last week, I've used the PWS functionality for the first time with real data.  I had some issues with our current PWS commands:
- The user has to provide the password as a CLI argument.  This might lead to leaks via the history file, logs or the process list.
- There are three typical use cases:
  1. Adding a new site.  Currently, I have to check which slots are currently programmed (`pws status`), keep track of the first unprogrammed slot and then call `pws set`.
  2. Updating existing information.  Currently, this requires me to re-enter all information for the `pws set` command instead of just listing the information to update.
  3. Deleting a site.  This works fine.

So I suggest to replace the current `pws set` command with:
- `pws set [--overwrite] <slot> <name> <login> [<password>|-]`.  If the password is set to `-`, it is read from stdin.  If it is not set, it is prompted from the user.  Unless `--overwrite` (or `--force`?) is set, this makes sure that the slot is currently unprogrammed to avoid accidental data loss.
- `pws add <name> <login> [<password>|-]`.  The same as `pws set`, but it queries the first unprogrammed slot and writes the data to this slot.
- `pws update <slot> [--name <name>] [--login <login>] [--password <password>]`.  Overwrites the given data on the given slot without changing the other fields.

If we want to be even more interactive, we could also make the `slot`, `name` and `login` arguments optional and prompt them from the user if not set.

@d-e-s-o What do you think?

d-e-s-o:
Hi @robinkrahl, your suggestions sound fine to me. It seems as if we should make `pws update --password` also work with stdin then, though, should we? I don't think we need to make input of `slot`, `name`, and `login` interactive.

robinkrahl:
> It seems as if we should make `pws update --password` also work with
> stdin then, though, should we?

True, good point!

d-e-s-o:
Is there some fundamental difference in OTP handling that we don't provide a similar interface there?

robinkrahl:
No, that would be my next step. I started with the PWS because the commands are more complex and data loss might hurt more.

d-e-s-o:
It seems as if this issue can be closed, correct? From what I can tell https://github.com/d-e-s-o/nitrocli/pull/163 was the last outstanding issue. Or is there something else missing, Robin?

robinkrahl:
I also intended to add a password prompt.  Just reading from stdin is fine for piping, but if the user wants to manually type the password, it’s a bit awkward.

PR #150, opened 2021-04-14
Add pws update subcommand

robinkrahl:
This patch adds the pws update subcommand that overwrites some fields of
a programmed PWS slot.  We always have to specify the values for all
fields when calling writing to a slot, so we first query the current
values of the fields that remain unchanged and then execute the write
command.  This is a potential race condition, but concurrent access to a
Nitrokey device is unsupported and highly likely to mess up the
communication anyway.

----

This is a first step towards implementing a better PWS interface as discussed in #139.


d-e-s-o:
Merged. Thanks for the pull request!

PR #151, opened 2021-04-14
Add pws add subcommand

robinkrahl:
This patch adds the pws add subcommand that locates the first
unprogrammed PWS slot and writes to it.

---

This patch is part of an effort to improve the PWS user interface, see #139.

This PR is based on #150.  (I thought it would be independent of it, but it made more sense to reuse the slot assertion helper method.)


d-e-s-o:
> This PR is based on #150. (I thought it would be independent of it, but it made more sense to reuse the slot assertion helper method.)

Beats me how a service/UI used by tens of thousand of people and being worked on full time doesn't have some kind of pull request stacking feature, to tell the truth ;-)

d-e-s-o:
Thanks for the change! I merged it into `devel`.

PR #153, opened 2021-04-16
Add GitHub Actions configuration

robinkrahl:
This patch adds a configuration for GitHub Actions that builds and tests
nitrocli with the MSRV, stable, beta and nightly Rust, calls clippy and
rustfmt and checks the license annotations with reuse-tool.

(Previous discussion at #148.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,68 @@
+# Copyright (C) 2021 The Nitrocli Developers
+# SPDX-License-Identifier: CC0-1.0
+
+# TODO:
+# - Test with system libnitrokey (USE_SYSTEM_LIBNITROKEY=1)?
+# - Add support for macos and windows
+
+name: CI
+
+on: [push, pull_request]
+
+env:
+  RUST_BACKTRACE: 1
+
+jobs:
+  test:
+    name: Compile and test Rust ${{ matrix.rust }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        rust: [1.42.0, stable, beta, nightly]
+    steps:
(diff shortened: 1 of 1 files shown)


robinkrahl:
[Build logs](https://github.com/robinkrahl/nitrocli/runs/2361872154) for this PR.

robinkrahl:
Notes:
- We might want to add caching using [`actions/cache`](https://github.com/actions/cache).
- Most examples use `cargo check` instead of `cargo build`. But as our build involves linking to a shared library (which AFAIK is skipped for `check`), I think performing the full build is better for us.

d-e-s-o:
Fabulous stuff! Thanks for looking into that.

> * We might want to add caching using [`actions/cache`](https://github.com/actions/cache).

We can do that. Let's first move forward with what we have. It seems to already be a hefty speed boost.

> * Most examples use `cargo check` instead of `cargo build`. But as our build involves linking to a shared library (which AFAIK is skipped for `check`), I think performing the full build is better for us.

:+1:

d-e-s-o:
Will merge tomorrow.

d-e-s-o:
Merged.

robinkrahl:
Thanks for merging!  By the way, I also [tested](https://github.com/robinkrahl/nitrocli/runs/2370546450?check_suite_focus=true) macOS, but we get an exception when trying to run the tests.

d-e-s-o:
> Thanks for merging! By the way, I also [tested](https://github.com/robinkrahl/nitrocli/runs/2370546450?check_suite_focus=true) macOS, but we get an exception when trying to run the tests.

Hm, interesting. No idea how to debug this issue, to be honest. We may still need someone with a macOS at hand.

PR #156, opened 2021-04-17
Add pws-cache extension

### LINKED ISSUE #155: Access PWS slots by name
Similar to #83, we also want to access PWS slots by name.  But we need the user PIN to read the slots (while we were able to read the OTP slots without authentication).  So this leaves us with these options:
1. Copy the `pinentry` module from `nitrocli` to `nitrocli-ext`, use it to query the user PIN and then use `nitrokey-rs` to access the PWS.
2. Same as 1., but move the `pinentry` module from `nitrocli` to `nitrocli-ext` (and make `nitrocli` depend on `nitrocli-ext`).
3. Call `nitrocli pws status` and parse its output.

@d-e-s-o What do you think?

robinkrahl:
This patch adds the pws-cache core extension that allows accessing the
PWS slots by their name instead of the slot index.

Fixes #155.

----

Open issues:  How to query the PWS data (requires user PIN).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased
 ----------
 - Introduced extension support crate, `nitrocli-ext`
-- Introduced `otp-cache` core extension
+- Introduced `otp-cache` and `pws-cache` core extensions
 - Enabled usage of empty PWS slot fields
 - Changed error reporting format to make up only a single line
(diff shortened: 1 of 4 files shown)


d-e-s-o:
Hi Robin, overall this pull request looks great, but I think we have to decide how to proceed with respect the PIN entry story that we discussed over in https://github.com/d-e-s-o/nitrocli/issues/155. Let me know how you want to proceed.

PR #157, opened 2021-04-17
Add overwrite option to pws set command

robinkrahl:
This patch adds the --overwrite option to the pws set command and
changes the default behavior to refuse overwriting programmed slots.

----

I originally suggested this change in #139 but I think I’ve changed my mind:  This form of the `pws set` command is pretty useless.  If I want to overwrite a slot I should rather use `pws update`.  If I want to add a new slot I most likely want to use `pws add`.  Only if I want to set a specific slot I need `pws set`.

At the same time, changing the default behavior is a already breaking change.  Why not remove `pws set` completely while at it and add a `--slot` option to `pws add` if I really want to write a specific slot?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Unreleased
 - Bumped `nitrokey` dependency to `0.9.0`
 - Added the `--only-aes-key` option to the `reset` command to build a new AES
   key without performing a factory reset
+- Added the `--overwrite` option to `pws set` and changed the default behavior
+  to not overwrite programmed slots
 
 
 0.4.0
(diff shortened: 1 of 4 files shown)


d-e-s-o:
Yeah, I agree. Let's just remove `pws set`. I like the `--slot` suggestion.

robinkrahl:
Do you want to remove `pws set` at once, or first deprecate it and remove it later?

d-e-s-o:
> Do you want to remove `pws set` at once, or first deprecate it and remove it later?

I don't think I necessarily kept track of every breaking change in the past and we already discussed that basically "anything goes" before 1.0. So I am fine with just removing it straight away.

PR #158, opened 2021-04-18
Add --slot option to pws add

robinkrahl:
This patch adds the --slot option the the pws add command so that we can
add data on a specific slot without using pws set (which might override
a programmed slot).

----

This patch is part of an effort to improve the PWS interface, see #139 for the tracking issue and #157 for additional context.


d-e-s-o:
Merged. Thanks Robin!

PR #160, opened 2021-04-19
Allow serial number filtering for Nitrokey Storage

d-e-s-o:
The Nitrokey Storage does not report a serial number in the USB device
descriptor. Because of that we have not supported filtering of those
devices by serial number. However, by actually connecting to the device
we can still retrieve such a number and, in fact, that's what we report
as part of the list command.
With this change we mirror this logic in the device discovery/filtering
pass: if the reported device information does not contain a serial
number (as is the case for Nitrokey Storage), we connect to the device
and look it up this way. In so doing we reduce the amount of confusion
surrounding the --serial-number option and configuration file argument.


d-e-s-o:
Hm. So one problem we have is that `nitrokey-test` does not have a mode that *just* serializes accesses but does not impose other constraints: we always require either at least one device to be present (if the test has a model/device argument) OR that there is none (if there is no filter and the test function has no argument).
Right now that's screwing us over because the extension tests are regular `#[test]`s that don't get serialized and so we run into concurrent access issues. But we have no good way to convert them straight over (or at least I don't see it).

d-e-s-o:
Making this a draft to discuss first. @robinkrahl, what do you think, is the approach viable? The serial restriction annoys me.

robinkrahl:
That’s a difficult problem.  I was hoping that a firmware patch would just make it disappear.  I’ll have to think about it a bit.

Anyway, I think connecting to a device should be the last resort.  We should check the serial numbers returned by the device enumeration before we start to connect to the other devices and query their serial numbers.

d-e-s-o:
> That’s a difficult problem.

Can you elaborate a bit, please?

> I was hoping that a firmware patch would just make it disappear. I’ll have to think about it a bit.

Yeah, that would be optimal, of course. If that's in the pipeline, that'd be great.

> Anyway, I think connecting to a device should be the last resort. We should check the serial numbers returned by the device enumeration before we start to connect to the other devices and query their serial numbers.

But isn't that what we do? In fact, we are already doing that as part of `list` (of course that's a different path). And I think that the discrepancy between reporting a serial number there but then not allowing its usage for the corresponding filtering is non-obvious to any end-user (including me, but then I recalled :D)

I feel like because that's all happening lazily and most users don't have multiple devices connected, there won't be any negative impact unless the feature is actually actively used (i.e., the user provided a serial number to filter by and a storage key is plugged in).

robinkrahl:
> > That’s a difficult problem.
> 
> Can you elaborate a bit, please?

Well, so far we’ve only connected to a single device.  The only exception was `list` but there we have a `--no-connect` option to disable the connections.  With this patch, I have to assume that `nitrocli` will connect to every attached Nitrokey device if I use the `--serial-number` option.  I agree that it is much more practical, but I dislike it conceptionally.  That’s the reason why I didn’t implement it when implementing the device selection logic.

> > I was hoping that a firmware patch would just make it disappear. I’ll have to think about it a bit.
> 
> Yeah, that would be optimal, of course. If that's in the pipeline, that'd be great.

I don’t think so.

> > Anyway, I think connecting to a device should be the last resort. We should check the serial numbers returned by the device enumeration before we start to connect to the other devices and query their serial numbers.
> 
> But isn't that what we do? In fact, we are already doing that as part of `list` (of course that's a different path). And I think that the discrepancy between reporting a serial number there but then not allowing its usage for the corresponding filtering is non-obvious to any end-user (including me, but then I recalled :D)
> 
> I feel like because that's all happening lazily and most users don't have multiple devices connected, there won't be any negative impact unless the feature is actually actively used (i.e., the user provided a serial number to filter by and a storage key is plugged in).

True, but I think there are at least two cases where we can skip the serial number queries:  If there are multiple devices that match based on the hidapi information, we already know that we return an error.  And if there is a single matching device and we are looking for exactly one serial number, we don’t have to check the serial numbers of the other devices.

> Hm. So one problem we have is that `nitrokey-test` does not have a mode that _just_ serializes accesses but does not impose other constraints: we always require either at least one device to be present (if the test has a model/device argument) OR that there is none (if there is no filter and the test function has no argument).
> Right now that's screwing us over because the extension tests are regular `#[test]`s that don't get serialized and so we run into concurrent access issues. But we have no good way to convert them straight over (or at least I don't see it).

As far as I see, the extensions don’t use the `--serial-number` option.  What is causing these issues?

Some thoughts regarding the tests in general:  In the future, I’d like to be able to run the tests concurrently on multiple devices.  So my suggestion is that `nitrokey-test` does no longer deal with the connection management.  Instead it only looks at `list_devices`, selects the appropriate devices and passes a `nitrokey::DeviceInfo`, an `Option<nitrokey::DeviceInfo>`, a `Vec<nitrokey::DeviceInfo>` or an `Option<Vec<nitrokey::DeviceInfo>>` to the test.  In the first two cases, the test would then only be allowed to connect to the USB paths of the given `DeviceInfo` instance.  `nitrokey-test` could use the USB paths for synchronization to make sure that there is no concurrent access.  The extension tests could then use `Option<nitrokey::DeviceInfo>` or `Option<Vec<nitrokey::DeviceInfo>>`.

d-e-s-o:
> > > Anyway, I think connecting to a device should be the last resort. We should check the serial numbers returned by the device enumeration before we start to connect to the other devices and query their serial numbers.
> > 
> > 
> > But isn't that what we do? In fact, we are already doing that as part of `list` (of course that's a different path). And I think that the discrepancy between reporting a serial number there but then not allowing its usage for the corresponding filtering is non-obvious to any end-user (including me, but then I recalled :D)
> > I feel like because that's all happening lazily and most users don't have multiple devices connected, there won't be any negative impact unless the feature is actually actively used (i.e., the user provided a serial number to filter by and a storage key is plugged in).
> 
> True, but I think there are at least two cases where we can skip the serial number queries: If there are multiple devices that match based on the hidapi information, we already know that we return an error. And if there is a single matching device and we are looking for exactly one serial number, we don’t have to check the serial numbers of the other devices.

Fair points. My thinking was that serial filtering alone is presumably already rare (although perhaps not so much with the configuration file), let alone in conjunction with multiple devices plugged in. So enablement was more important than optimizing for all possible cases for me. We can see if it's possible to cater to these cases nicely.

> > Hm. So one problem we have is that `nitrokey-test` does not have a mode that _just_ serializes accesses but does not impose other constraints: we always require either at least one device to be present (if the test has a model/device argument) OR that there is none (if there is no filter and the test function has no argument).
> > Right now that's screwing us over because the extension tests are regular `#[test]`s that don't get serialized and so we run into concurrent access issues. But we have no good way to convert them straight over (or at least I don't see it).
> 
> As far as I see, the extensions don’t use the `--serial-number` option. What is causing these issues?

It has to do with the `Manager` acquisition that we are now doing on that path. So it does not matter if they use `--serial-number`. Our extension tests are not `nitrokey-test` tests and so they run concurrently (unless explicitly restricted through `cargo test`). What can happen is that two attempt to `take` the manager at the same time. But I see I could just use `take_blocking` to circumvent that. Not pretty, though.

> Some thoughts regarding the tests in general: In the future, I’d like to be able to run the tests concurrently on multiple devices. So my suggestion is that `nitrokey-test` does no longer deal with the connection management. Instead it only looks at `list_devices`, selects the appropriate devices and passes a `nitrokey::DeviceInfo`, an `Option<nitrokey::DeviceInfo>`, a `Vec<nitrokey::DeviceInfo>` or an `Option<Vec<nitrokey::DeviceInfo>>` to the test. In the first two cases, the test would then only be allowed to connect to the USB paths of the given `DeviceInfo` instance. `nitrokey-test` could use the USB paths for synchronization to make sure that there is no concurrent access. The extension tests could then use `Option<nitrokey::DeviceInfo>` or `Option<Vec<nitrokey::DeviceInfo>>`.

I think that makes sense. It would sort of disambiguate the case where multiple keys of the same model are plugged in (that I was worried about recently), but at the same time we would need to define the semantics of that exactly (do we just go round-robin on devices of the same model? that may or may not be desirable: faster, but they could differ in terms of firmware). We'd also need to make sure that we keep boilerplate in the tests to a minimum.

robinkrahl:
> My thinking was that serial filtering alone is presumably already rare (although perhaps not so much with the configuration file), let alone in conjunction with multiple devices plugged in.

On the other hand, if multiple devices are plugged in, filtering by serial number becomes much more relevant.  Anyway, I think we will find a solution that enables the `--serial-number` argument for the Storage and minimizes the side effects.

> It has to do with the `Manager` acquisition that we are now doing on that path.

I see, thanks for the explanation.

> It would sort of disambiguate the case where multiple keys of the same model are plugged in (that I was worried about recently), but at the same time we would need to define the semantics of that exactly (do we just go round-robin on devices of the same model? that may or may not be desirable: faster, but they could differ in terms of firmware).

We could think about two modes:  a fast mode that chooses a random matching device that we can use for development, and an extensive mode that runs every test on every matching device that we can run before a release or after major changes.  Some amount of randomization might be useful to detect implicit dependencies between the tests.

robinkrahl:
I just stumbled over https://github.com/Nitrokey/nitrokey-storage-firmware/issues/76.  Could you please compare the `gpg --card-status` and `nitrocli status` output for your Nitrokey Storage?  Do they match?  Does `nitrocli status` report the same serial number as it did for my Storage device?

d-e-s-o:
I am seeing different serial numbers between `gpg --card-status` and `nitrocli status`. The serial number that `nitrocli status` reports is the same that you reported (`00050000`; d'oh).

robinkrahl:
Thanks.  In this case, it seems like serial number filtering for Nitrokey Storage would only act as syntactic sugar for `--model storage`, right?

d-e-s-o:
Yeah, doesn't make sense then. Sigh.

PR #162, opened 2021-04-20
Remove pws set

robinkrahl:
This patch removes the pws set subcommand as part of the PWS interface
refactoring, see #139.  Users should use pws add (for writing to new slots) or pws
update (for writing to existing slots) instead.  This makes it less
likely to accidentally overwrite PWS data.

### DIFF
--- a/src/tests/pws.rs
+++ b/src/tests/pws.rs
@@ -5,6 +5,24 @@
 
 use super::*;
 
+fn clear_pws(model: nitrokey::Model) -> anyhow::Result<()> {
+  use nitrokey::GetPasswordSafe as _;
+
+  let mut manager = nitrokey::force_take()?;
+  let mut device = manager.connect_model(model)?;
+  let mut pws = device.get_password_safe(nitrokey::DEFAULT_USER_PIN)?;
+  let slots_to_clear: Vec<_> = pws
+    .get_slots()?
+    .into_iter()
+    .flatten()
+    .map(|s| s.index())
+    .collect();
+  for slot in slots_to_clear {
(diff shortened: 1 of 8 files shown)


src/tests/pws.rs
```diff
@@ -130,8 +146,11 @@ fn set_reset_get(model: nitrokey::Model) -> anyhow::Result<()> {
 
 #[test_device]
 fn clear(model: nitrokey::Model) -> anyhow::Result<()> {
+  clear_pws(model)?;
+
   let mut ncli = Nitrocli::new().model(model);
-  let _ = ncli.handle(&["pws", "set", "10", "clear-test", "some-login", "abcdef"])?;
+  let _ = ncli.handle(&["pws", "clear", "10"])?;
```
d-e-s-o, line 152: Hm, why is that call needed?

d-e-s-o:
Merged! Thank you Robin! FYI: I removed the `pws clear 10` call that I didn't understand why we would need. Let me know if it had some deeper intention that I just didn't grok.

d-e-s-o:
Oh, and I didn't revert the MSRV bump. I honestly think it's fine to go a bit higher in the supported Rust version, so that we can use newer features over time; and `1.43.0` is already a year old by now. Let me know if there are some requirements to stay at 1.42 for packaging or something like that.

src/tests/pws.rs
robinkrahl: It’s not needed, but I figured it would be good to check that we can clear an unprogrammed slot without errors.  But we can also move that into a separate test case.

PR #163, opened 2021-04-20
Support reading PWS passwords from stdin

robinkrahl:
Specifying passwords as command-line arguments can be problematic:  The
user has to be careful that the command is not recorded in the shell
history.  Also, the password might be exposed over the process list.

This patch adds an alternative input method for the pws add and pws
update methods: reading the password from stdin.  This is activated by
setting the password to "-".  For the tests, we use a string as a
simulated stdin buffer.

----

This patch is part of the PWS interface refactoring discussed in #139 and is based on #162.


d-e-s-o:
Very nice change. Thanks Robin! Merged.

Issue #166, opened 2021-04-23
Publishing the core extensions


robinkrahl:
How should the core extensions be published?  In my opinion, they should be published the same way as `nitrocli`, both for usability and maintainability reasons.  So I see three options:

- Ditch crates.io for all our releases.  We should have good reasons for this step, and I don’t see any.
- Publish `nitrocli` and all core extensions as separate crates on crates.io.  This is how I understand our current approach.
- Have the core extensions as additional binary targets for the `nitrocli` crate (maybe behind feature flags) and release them together with `nitrocli`.  This might be easier to maintain than separate crates, but then we wouldn’t really need the extension mechanism any more.

d-e-s-o:
So far my thinking was to not publish them to `crates.io`. We'd still be publishing `nitrocli`, though. `nitrocli` has started off being published there and it is useful by itself.

Extensions change the picture: for one, they (generally) are not useful without `nitrocli`. So now there is a dependency. They are also not entirely trivial to make work with the main program: you need to install them at a proper location. In my opinion that is the job of some kind of distribution package, to take care of all of that.

I understand (hope?) that all Linux distributions that we support build binary packages from source. And by that I mean the actual repository contents, not the archive uploaded to `crates.io`. So nothing changes for them and they don't need anything published to `crates.io`.
I don't know about Mac & Windows, but I am also reluctant to call them supported to begin with. In any case, the option to build from source is always available and just having something on `crates.io` in my opinion only helps marginally.

Of course, publishing extensions there also does not hurt much. And I am not fundamentally against publishing them. I just think that given that there is more involved to getting them to work than what `cargo` can do out of the box, taking care of having proper packages seems more important. So from my perspective, publishing of extensions is not necessary but we can go down that route if there is a convincing use case.

d-e-s-o:
Do you have concrete use cases for publishing to `crates.io` or is that more a "we should do it because it seems to be the right thing to do" kind of stance, @robinkrahl? As you can probably tell, I am much less opposed to publishing core extensions than I am to publishing `nitrocli-ext`, so I am trying to get a better picture here.

robinkrahl:
> I understand (hope?) that all Linux distributions that we support build binary packages from source. And by that I mean the actual repository contents, not the archive uploaded to `crates.io`.

The Debian packages are built from the releases published to crates.io.

> They are also not entirely trivial to make work with the main program: you need to install them at a proper location.

[...]

PR #169, opened 2021-04-25
Make otp-cache a binary target of nitrocli

robinkrahl:
This is intended as an example for discussion in #166.  Unforunately, it
is not possible to directly include modules from the main binary, so we
use
```
    #[path = "../ext.rs"]
    mod ext;
```
as a workaround to share code between the extensions.


d-e-s-o:
Thanks for making this change! Looks good to me. The only thing that I think we should change is to keep all things extensions  below `ext/`. This makes the rather strict separation from the main program more clear. I'd want to reserve `src/` strictly for `nitrocli` proper (yes, it's a bit of a misnomer then but let's accept that for now; I don't believe that we are alone in this boat).

If that makes sharing code between extensions and `nitrocli` harder then I also think that's a good thing. We should be very deliberate about that.

robinkrahl:
> The only thing that I think we should change is to keep all things extensions below `ext/`.

What about the `ext` module?  Should it be `ext/mod.rs` or `src/ext.rs`?

d-e-s-o:
> > The only thing that I think we should change is to keep all things extensions below `ext/`.
> 
> What about the `ext` module? Should it be `ext/mod.rs` or `src/ext.rs`?

I'd just go with `<nitrocli-root>/ext/ext.rs`. If we decide to go multi-file we'd make that a directory `<nitrocli-root>/ext/ext/{mod,impl}.rs`. I suppose that should "just work" then. We can also do that right away if you feel that's a distinct possibility.

robinkrahl:
> I'd just go with `<nitrocli-root>/ext/ext.rs`.

Okay!  Though I was just happy to get rid of the `ext/ext` duplication in the path. ;)

d-e-s-o:
Shall we merge this pull request, Robin? I also think we can remove `Context::text` for now if it's not actively used (but I can take care of that).

robinkrahl:
If we want to use this approach, sure.  I’ve removed the Nitrocli::text
method, but we’ll probably have to add #[allow(unused)] in the future
anyway because not all extensions will use all parts of the interface.

d-e-s-o:
I merged the change. Thanks Robin!

PR #170, opened 2021-04-25
Improve error messages for overlong input data

### LINKED ISSUE #161: Validate PWS and OTP string length
Currently, we leave the validation of PWS slot names, logins and passwords as well as OTP slot names and secrets to `libnitrokey` which returns an error if a string is too long.  There are two problems with this:
1. Slot names with length > 15 and <= 30 are truncated, see https://github.com/Nitrokey/libnitrokey/issues/196.
2. It is not possible to see which string is too long.  This is especially bad for the PWS with three potential offenders (and three different length limits).

While I consider 1. to be a bug in `libnitrokey` and/or the firmware, 2. is a legitimate issue.  I suggest to store the maximum lengths as constants in `nitrocli` and to validate the user input, providing helpful error messages like:

```
The name and the login for the PWS slot are too long (maximum length: x bytes for the name, y bytes for the login).
```

@d-e-s-o What do you think?

robinkrahl:
Previously, we didn’t check the length of the input data in the otp set,
pws add and pws update commands.  While libnitrokey does check the input
and return an error if a string is too long, it cannot inform us which
of the input strings is the problematic one.

With this patch, we manually validate the input string lengths for these
commands to improve the error messages, clearly stating which strings
are problematic, how long they are and how long they can be.

Fixes #161.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Unreleased
 - Added the `--only-aes-key` option to the `reset` command to build a new AES
   key without performing a factory reset
 - Added support for reading PWS passwords and OTP secrets from stdin
+- Changed the `otp set`, `pws add` and `pws update` commands to check the
+  length of the input data to improve the error messages
 - Added `NITROCLI_RESOLVED_USB_PATH` environment variable to be used by
   extensions
 - Allowed entering of `base32` encoded strings containing spaces
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -35,6 +35,12 @@ use crate::Context;
 
 const NITROCLI_EXT_PREFIX: &str = "nitrocli-";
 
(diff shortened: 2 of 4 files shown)


d-e-s-o:
Great work! Thanks Robin! Should we have an issue open and a TODO referencing it to not forget about the move away from hardcoding these maximum lengths in `nitrocli`?

robinkrahl:
Thanks!  We have the libnitrokey issue and PR.  Once a new libnitrokey version is released, I’ll update nitrokey-sys.  And for every nitrokey-sys update, I check which changes are relevant for nitrocli.  So having another issue or TODO is not necessary for me, but feel free to add it if you want to.

d-e-s-o:
Sounds good. That should be good enough then. Merged!

PR #174, opened 2021-05-02
Refactor status to avoid unnecessary commands

robinkrahl:
I wrote the first version of the current status command when I was not
too familiar with libnitrokey inner workings.  This resulted in a
combinaton of funtions that requires more command invocations than
necessary.  With this patch, we can reduce the number of command
invocations from three to one GetStatus commands for the Nitrokey Pro
and from six to two GetDeviceStatus commands for the Nitrokey Storage.

Once a minor issue in nitrokey-rs [0] is fixed, we can further reduce
the Nitrokey Storage case to a single GetDeviceStatus invocation.

[0] https://todo.sr.ht/~ireas/nitrokey-rs/1

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -433,12 +433,15 @@ fn ensure_string_lengths(data: &[(&str, &str, usize)]) -> anyhow::Result<()> {
   }
 }
 
-/// Query and pretty print the status that is common to all Nitrokey devices.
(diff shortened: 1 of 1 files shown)


d-e-s-o:
Good stuff! Merged. Thanks Robin!

PR #178, opened 2021-05-02
Fix handling of empty user input in pinentry module

### LINKED ISSUE #176: "Unexpected response: OK" if empty password is entred via pinentry
Hi!

During investigation of issue #175 I found that nitrocli-0.4.0 fails to recognize pinentry's response when empty password is provided and OK button is pressed (or just hit the Enter):
```
$  nitrocli pws get 1
Failed to parse pinentry secret

Caused by:
    Unexpected response: OK
```
If cancel is pressed (or Esc) response is more reasonable:
```
$  nitrocli pws get 1
Failed to parse pinentry secret

Caused by:
    83886179 Operation cancelled <Pinentry>
```
Why I need to enter empty password in the first place? Due to bug #175 the first entered password will be discarded regardless of whether it is valid or not, so to save time I tried just to skip it by pressing Enter and found another issue.

d-e-s-o:
As it turns out, the pinentry program responds slightly differently to a
request to enter a password/PIN when the user enters one vs. when he/she
just continues, leaving the input empty, as is illustrated here:

When the user enters 'hello':
  > $ gpg-connect-agent 'GET_PASSPHRASE --data cacheid errormsg prompt description' '/bye'
  > D hello
  > OK

When the user enters nothing:
  > $ gpg-connect-agent 'GET_PASSPHRASE --data cacheid errormsg prompt description' '/bye'
  > OK

So far we always expected a two line response, with the first line
containing the 'D ' prefix, which prevents us from handling empty input
properly.
This change fixes this shortcoming by adding a case for it.

Fixes #176

### DIFF
--- a/src/pinentry.rs
+++ b/src/pinentry.rs
@@ -204,7 +204,7 @@ where
     ["OK"] => Ok(String::new()),
     [line, "OK"] if line.starts_with(OK_PREFIX) => {
       let (_, pass) = line.split_at(OK_PREFIX.len());
-      return Ok(pass.to_string());
+      Ok(pass.to_string())
     }
     [line] if line.starts_with(ERR_PREFIX) => {
       let (_, error) = line.split_at(ERR_PREFIX.len());
--- a/src/pinentry.rs
+++ b/src/pinentry.rs
@@ -1,6 +1,6 @@
 // pinentry.rs
 
-// Copyright (C) 2017-2020 The Nitrocli Developers
+// Copyright (C) 2017-2021 The Nitrocli Developers
 // SPDX-License-Identifier: GPL-3.0-or-later
 
 use std::borrow;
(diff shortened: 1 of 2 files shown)


src/pinentry.rs
```diff
@@ -187,6 +187,9 @@ fn parse_pinentry_pin<R>(response: R) -> anyhow::Result<String>
 where
   R: AsRef<str>,
 {
+  const OK_PREFIX: &str = "D ";
```
robinkrahl, line 190: Intuitively, I expected `OK_PREFIX` to be `"OK"`.  Maybe `DATA_PREFIX`?

robinkrahl (approved):
Thanks, looks good to me!

src/pinentry.rs
d-e-s-o: Good call. Will change it. Thanks for your review!

PR #180, opened 2021-07-30
Update Arch Linux package documentation

robinkrahl:
The Arch Linux package is now available from the official Arch
repositories, so we no longer need the AUR-specific documentation.

### DIFF
--- a/doc/packaging.md
+++ b/doc/packaging.md
@@ -6,24 +6,7 @@ This document describes how to update the packaged versions of nitrocli.
 Arch Linux
 ----------
 
-1. Clone the Git repository at ssh://aur@aur.archlinux.org/nitrocli.git.
-2. Edit the `PKGBUILD` file:
-   - Update the `pkgver` variable to the current nitrocli version.
-   - If the `pkgrel` variable is not 1, set it to 1.
-   - Update the SHA512 hash in the `sha512sums` variable for the new tarball.
-3. Update the `.SRCINFO` file by running `makepkg --printsrcinfo > .SRCINFO`.
-4. Verify that the package builds successfully by running `makepkg`.
-5. Verify that the package was built as expected by running `pacman -Qlp $f`
-   and `pacman -Qip $f`, where `$f` is `nitrocli-$pkgver.pkg.tar.gz`.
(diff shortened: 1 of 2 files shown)


doc/packaging.md
```diff
@@ -6,24 +6,7 @@ This document describes how to update the packaged versions of nitrocli.
...
-6. Check the package for errors by running `namcap PKGBUILD` and `namcap
-   nitrocli-$pkgver.pkg.tar.gz`.
-7. Add, commit, and push your changes to publish them in the AUR.
-
-If you have to release a new package version without a new nitrocli version,
-do not change the `pkgver` variable and instead increment the `pkgrel`
-variable.
-
-For more information, see the [Arch User Repository][] page in the Arch Wiki.
+The Arch Linux package is now maintained as part of the community repository.
```
d-e-s-o, line 9: I'd remove the "now". Shouldn't really matter what the past looked like for the purposes of this document.

d-e-s-o (commented):
Great stuff Robin! Thanks!

One question: so how does the workflow work now, precisely, from the Arch side? Does Maxime Gauduin monitor new releases being tagged here? Or how else is he made aware, if there is no action required from our side?

doc/packaging.md
robinkrahl: Good point.  Can you just remove it while merging?

robinkrahl:
> One question: so how does the workflow work now, precisely, from the Arch side? Does Maxime Gauduin monitor new releases being tagged here? Or how else is he made aware, if there is no action required from our side?

I assume he is using GitHub notifications or RSS feeds.  @alucryd Can you maybe help us here?  (And thanks for taking over the package!)

alucryd:
@robinkrahl No problem! I have a buildbot that automatically builds new git tags. Time to repo should be fairly quick as I just need to do a quick manual check before committing the new version to our repo.

robinkrahl:
@alucryd Nice!  Thanks for the explanation!

doc/packaging.md
d-e-s-o: Absolutely.

d-e-s-o:
Merged. Thanks Robin & Maxime!

PR #184, opened 2021-12-10
Migrate to clap 3.0.0

robinkrahl:
Current status:
- 2021-12-10: compiles with clap 3.0.0 and should already work quite well

To do:
- [ ] double-check extension handling
- [ ] check version output and test case
- [ ] update help text tests
    - [ ] remove help text tests?
- [ ] use `clap::ArgEnum`
- [ ] replace `StructOpt`/`structopt` with `Parser`/`clap`
- [ ] re-evaluate overall argument handling with regard to the new clap API
- [ ] consider making `clap_generate` optional
- [ ] re-evaluate `clap` feature selection


d-e-s-o:
Thanks for the pull request! Will take a look this weekend.

d-e-s-o:
Looks good so far, @robinkrahl. It seems we may have to bump the minimum supported Rust version to at least 1.51 with these changes.

robinkrahl:
Thanks for the review!  Do you have an opinion on whether we should keep
the help text tests after the migration?  I’m still undecided.  They
might be a good reminder for updating the man page if the CLI changes,
but they also create some noise.

d-e-s-o:
> Do you have an opinion on whether we should keep the help text tests after the migration?

It's not a particularly strongly held opinion, but I'd remove them. As you say, there may some value as a reminder, but ultimately I suspect that will only work until we are conditioned to accepting changes in those tests without questioning them much. I also believe we should be able to get similar value through peer reviews.

PR #185, opened 2021-12-25
Include git commit hash in version output

d-e-s-o:
To better identify issues or just get an idea which change exactly a
given binary was built from, this patch makes sure to contain the Git
commit hash in the version output we provide both in the help text but
also via the -V/--version flag.

Note that I did not adjust the currently ignored `version_option` test. We will have to fix it once we migrate to clap 3.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,115 @@
+// Copyright (C) 2021-2022 The Nitrocli Developers
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+use std::process::Command;
+
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+
+const GIT: &str = "git";
+
+/// Format a git command with the given list of arguments as a string.
+fn git_command(args: &[&str]) -> String {
+  args.iter().fold(GIT.to_string(), |mut cmd, arg| {
+    cmd += " ";
+    cmd += arg;
+    cmd
+  })
+}
+
+/// Run git with the provided arguments and read the output it emits.
+fn git_output(args: &[&str]) -> Result<String> {
+  let git = Command::new(GIT)
+    .args(args)
+    .output()
+    .with_context(|| format!("failed to run `{}`", git_command(args)))?;
+
+  if !git.status.success() {
+    let code = if let Some(code) = git.status.code() {
+      format!(" ({})", code)
+    } else {
+      String::new()
+    };
+
+    bail!(
+      "`{}` reported non-zero exit-status{}",
+      git_command(args),
+      code
+    );
+  }
+
+  let output = String::from_utf8(git.stdout).with_context(|| {
+    format!(
+      "failed to read `{}` output as UTF-8 string",
+      git_command(args)
+    )
+  })?;
+
+  Ok(output)
+}
+
+/// Run git with the provided arguments and report the status of the
+/// command.
+fn git_run(args: &[&str]) -> Result<bool> {
+  Command::new(GIT)
+    .args(args)
+    .status()
+    .with_context(|| format!("failed to run `{}`", git_command(args)))
+    .map(|status| status.success())
+}
+
+/// Retrieve a git revision identifier that either includes the tag we
+/// are on or the shortened SHA-1. It also contains an indication
+/// whether local changes were present.
+fn git_revision() -> Result<Option<String>> {
+  // As a first step we check whether we are in a git repository and
+  // whether git is working to begin with. If not, we can't do much; yet
+  // we still want to allow the build to continue, so we merely print a
+  // warning and continue without a git revision. But once these checks
+  // are through, we treat subsequent failures as unexpected and fatal.
+  match git_run(&["rev-parse", "--git-dir"]) {
+    Ok(true) => (),
+    Ok(false) => {
+      println!("cargo:warning=Not in a git repository; unable to embed git revision");
+      return Ok(None);
+    }
+    Err(err) => {
+      println!(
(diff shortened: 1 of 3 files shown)


d-e-s-o:
@robinkrahl what do you think of the following change?

robinkrahl:
Thanks for the PR!  I’ll have a closer look at this in the next days,
but some thoughts off the top of my head:

1. I’d prefer to omit the Git version if we’re on the tag for the crate
   version.  This would make the output more concise and it would be
   clear at first glance whether a release is used or not.
2. If I understand this correctly, the changes flag will be wrong if I
   build, change a file and then build again.  What do you think about
   changing `rerun-if-changed=.git/index` to `rerun-if-changed=.` to be
   on the safe side?  It should not be too expensive to execute the
   build script quite often.

d-e-s-o:
Thanks for the review!

> 1. I’d prefer to omit the Git version if we’re on the tag for the crate version. This would make the output more concise and it would be clear at first glance whether a release is used or not.

That makes sense. Done. (though if there is a more elegant way feel free to suggest it)

> 2. If I understand this correctly, the changes flag will be wrong if I build, change a file and then build again. What do you think about changing `rerun-if-changed=.git/index` to `rerun-if-changed=.` to be on the safe side? It should not be too expensive to execute the build script quite often.

Yeah, that's a good catch. I forgot about that when I added the tracking of local changes. I am a bit wary of completely preventing no-op incremental rebuilds from happening (which as per my experimentation at least is what your suggestion amounts to [though I am not sure why exactly that's the case]). I am now only tracking the source directories and `.git` for changes. That's hopefully still maintainable enough.

robinkrahl:
Thanks for the update, looks good.  Can you also add Cargo.{lock,toml}
and build.rs itself to the file list?  That should catch all files that
directly affect the build process.

One nit regarding the current implementation:  I find the usage of Cow
to store an empty value counter-intuitive.  I think it would be fair to
use String::new instead because it does not allocate and is generally
inexpensive.  If you want to make it more explicit, I’d prefer an
Option<String> and .as_deref().unwrap_or_default() to access the value.

robinkrahl:
I forgot to bring up another point:  This breaks the build if it is not
run from a Git working tree, right?  There are at least two cases that
we should keep in mind:  the release tarballs and the version published
on crates.io which is used by cargo install.

We could support these cases by just omitting the Git suffix if the
commands fail.

d-e-s-o:
> Thanks for the update, looks good. Can you also add Cargo.{lock,toml} and build.rs itself to the file list? That should catch all files that directly affect the build process. One nit regarding the current implementation: I find the usage of Cow to store an empty value counter-intuitive. I think it would be fair to use String::new instead because it does not allocate and is generally inexpensive. If you want to make it more explicit, I’d prefer an Option<String> and .as_deref().unwrap_or_default() to access the value.

All addressed. `build.rs` is already tracked by Cargo. I had to get back to poking into `.git/`, though, because Cargo was modifying something inside `.git/` after executing `build.rs`, which was a/the source of constant rebuilds.

d-e-s-o:
> I forgot to bring up another point: This breaks the build if it is not run from a Git working tree, right? There are at least two cases that we should keep in mind: the release tarballs and the version published on crates.io which is used by cargo install. We could support these cases by just omitting the Git suffix if the commands fail.

True, that may cause trouble. Shall we go back to emitting the revision unconditionally then, though? Otherwise we have no way to distinguish between on-a-tag and no-longer-containing-.git. And I think that defeats the purpose of the mechanism to begin with.

d-e-s-o:
Will you be able to take a look, @robinkrahl? Otherwise I'll go ahead and merge the change to not have too many loose ends around.

robinkrahl:
I’ll try to review tomorrow or on Friday.  Sorry for the delay.  Anyway,
I still think it would be good to clearly identify versions built from a
release tag, even if we don’t drop the Git suffix entirely.

build.rs
```diff
@@ -0,0 +1,111 @@
...
+        err
+      );
+      return Ok(String::new());
+    }
+  }
+
+  let local_changes = git_output(&["status", "--porcelain", "--untracked-files=no"])?;
+  let modified = !local_changes.is_empty();
+  let revision = git_output(&["rev-parse", "--short", "HEAD"])?;
+  let suffix = format!(" ({}{})", revision.trim(), if modified { "+" } else { "" });
```
robinkrahl, line 89: I’d like to have an indicator for builds from a tag.  Otherwise I always have to look up the revision for a tag to determine whether a release was built.  I could think of these options:
- If `git describe --exact-match` equals `v${CARGO_PKG_VERSION}`, use `"tag"` instead of `revision`:
  ```
  nitrocli 0.4.1 (tag+) using libnitrokey v3.6
  ```
- Use `git describe --always` instead of `git rev-parse --short HEAD`:
    ```
  nitrocli 0.4.1 (v0.4.1) using libnitrokey v3.6
  nitrocli 0.4.1 (v0.4.1-12-gad55ab2) using libnitrokey v3.6
  ```

robinkrahl (commented):
I think I’d prefer a different arrangement of the version components to make it clear what the Git revision means, for example:

```
nitrocli 0.4.1 using libnitrokey v3.6 (built from <revision>)
```

Otherwise, the code looks good to me.

build.rs
d-e-s-o: I am embedding the tag name now, if there is one. I think that's the most readable and useful.

d-e-s-o:
Thanks for your review! I addressed all comments and merged the change.

Issue #2, opened 2018-05-20
Support for OTPs on the Nitrokey Pro?


robinkrahl:
I’m about to write a command-line tool to generate one-time passwords on my Nitrokey Pro. I already wrote a Rust binding for `libnitrokey` (crate [`nitrokey`][]). Before I create yet another Nitrokey CLI tool: Would you be interested in pull requests adding support for one-time password using libnitrokey? (I’m not keen on dealing with the raw hidapi communication, so that would not be an option for me.)

[`nitrokey`]: https://crates.io/crates/nitrokey

d-e-s-o:
Hi Robin. In general the answer is yes, I am open to and interested in combining efforts. My main worry is that we cannot have both: raw hidapi style communication as well as integration with `libnitrokey`. It appears `nitrokey` would not quite be capable of backing the existing functionality as it stands. Do you plan on adding support to the point that `nitrocli` could fully migrate over?

robinkrahl:
Good to hear! I’m happy to extend the `nitrokey` crate, but as I do not have a Nitrokey Storage, I cannot test anything related to it. Would you be willing to run the tests?

As far as I see, you are using these commands in `nitrocli`:
- `EnableEncryptedVolume` (0x20)
- `DisableEncryptedVolume` (0x21)
- `GetDeviceStatus` (0x2E)

The first two correspond directly to `NK_unlock_encrypted_volume` and `NK_lock_encrypted_volume` in `libnitrokey`. I can easily add these to the `nitrokey` crate. `GetDeviceStatus` is more problematic. While the C++ API provides access to the complete status, the C API only provides access to a string version (`NK_get_status_storage_as_string`). We can get the firmware version and the retry counts from other sources (already available in the `nitrokey` crate), but not the volume and SD card information.

So I am going to add support for enabling and disabling encrypted volumes to the `nitrokey-rs`, and I’ll prepare a patch for `libnitrokey` to add the full device status to the C API.

jans23:
@robinkrahl In case you are interested in adding full support for Nitrokey Storage, we would be glad to donate you a device.

robinkrahl:
@jans23 That would be great! I can’t make promises regarding the timeline, but I’m definitely interested in porting all `libnitrokey` features to Rust and, eventually and if applicable, to a command-line application.

d-e-s-o:
> Would you be willing to run the tests?

[...]

PR #191, opened 2022-09-16
Add Nix flake

VTimofeenko:
This commit embeds a Nix flake into the repository allowing a Nix user to run nitrocli by issuing:

    nix run d-e-s-o/nitrocli

This also adds a reproducible way to install nitrocli on a system controlled by Nix flakes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,39 @@ $ cargo build --release
 It is recommended that the resulting executable be installed in a
 directory accessible via the `PATH` environment variable.
 
+#### With Nix flakes
+##### Running nitrocli
+
+Repository comes with a `flake.nix` file, so it can be run directly:
+
+``` sh
+$ nix run d-e-s-o/nitrocli
+```
+##### Installing system-wide
+
+`nitrocli` can be installed by adding the repository flake as an input:
+
+``` nix
+{
+  inputs = {
+    nitrocli.url = "github:d-e-s-o/nitrocli?dir=contrib/nix";
+    ...
+  };
+
+  outputs = {
+    nitrocli,
+    ...
+  }:
+  {
+    # ...
+    # Where modules are defined
+    environment.systemPackages = [ nitrocli.defaultPackage ];
+    };
+    # ...
+}
+```
+
+
 
 #### Shell Completion
 **nitrocli** comes with completion support for options and arguments to
--- a/contrib/nix/flake.nix
+++ b/contrib/nix/flake.nix
@@ -1,0 +1,44 @@
+# Copyright (C) 2020 The Nitrocli Developers
+# SPDX-License-Identifier: CC0-1.0
+{
+  inputs = {
+    naersk.url = "github:nix-community/naersk/master";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, utils, naersk }:
+    utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs { inherit system; };
+        naersk-lib = pkgs.callPackage naersk { };
+      in
+      rec
+      {
+        packages.default = naersk-lib.buildPackage {
+          root = ./../../.;
+          nativeBuildInputs = with pkgs; [ hidapi ];
+          postInstall = ''
+            # copy the manpages
+            install -D --mode 0644 ${./../../.}/doc/nitrocli.1 $out/share/man/man1
+            # make completions
+            mkdir --parents $out/share/bash-completion/completions/
+            cargo run --bin=shell-complete bash > $out/share/bash-completion/completions/nitrocli
+            mkdir --parents $out/share/zsh/site-functions/
+            cargo run --bin=shell-complete zsh > $out/share/zsh/site-functions/_nitrocli
+            mkdir --parents $out/share/fish/vendor_completions.d/
+            cargo run --bin=shell-complete fish > $out/share/fish/vendor_completions.d/nitrocli.fish
+          '';
+        };
+
+        apps.default = utils.lib.mkApp {
+          drv = packages.default;
+          name = "nitrocli";
+        };
+
+        devShell = with pkgs; mkShell {
+          buildInputs = [ cargo rustc rustfmt pre-commit rustPackages.clippy hidapi ];
+          RUST_SRC_PATH = rustPlatform.rustLibSrc;
+        };
+      });
+}
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,0 +1,6 @@
+This is the `contrib/` directory of nitorcli project. Files located here:
+
+* Are provided by the community of the project
+* Are not official part of the project
+* Are provided without any warranty
+
--- a/contrib/README.md.license
+++ b/contrib/README.md.license
@@ -1,0 +1,2 @@
+Copyright (C) 2020 The Nitrocli Developers
+SPDX-License-Identifier: CC0-1.0
--- a/contrib/nix/flake.lock
+++ b/contrib/nix/flake.lock
@@ -1,0 +1,77 @@
+{
+  "nodes": {
+    "naersk": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      },
+      "locked": {
+        "lastModified": 1662220400,
+        "narHash": "sha256-9o2OGQqu4xyLZP9K6kNe1pTHnyPz0Wr3raGYnr9AIgY=",
+        "owner": "nix-community",
+        "repo": "naersk",
+        "rev": "6944160c19cb591eb85bbf9b2f2768a935623ed3",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "ref": "master",
+        "repo": "naersk",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1663264531,
+        "narHash": "sha256-2ncO5chPXlTxaebDlhx7MhL0gOEIWxzSyfsl0r0hxQk=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "454887a35de6317a30be284e8adc2d2f6d8a07c4",
+        "type": "github"
+      },
+      "original": {
+        "id": "nixpkgs",
+        "type": "indirect"
+      }
+    },
+    "nixpkgs_2": {
+      "locked": {
+        "lastModified": 1663264531,
+        "narHash": "sha256-2ncO5chPXlTxaebDlhx7MhL0gOEIWxzSyfsl0r0hxQk=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "454887a35de6317a30be284e8adc2d2f6d8a07c4",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixpkgs-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "naersk": "naersk",
+        "nixpkgs": "nixpkgs_2",
+        "utils": "utils"
+      }
+    },
+    "utils": {
+      "locked": {
+        "lastModified": 1659877975,
+        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "c0e246b9b83f637f4681389ecabcb2681b4f3af0",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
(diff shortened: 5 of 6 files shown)


robinkrahl:
Thanks for the PR!  Is it possible to add the man page and the shell completions to the flake?

VTimofeenko:
Sure. Would you prefer that to be in the same commit or a separate one?

d-e-s-o:
Apologies for the delay. I will should be able to take a closer look by the end of the weekend.

README.md
```diff
@@ -112,6 +112,39 @@ $ cargo build --release
 It is recommended that the resulting executable be installed in a
 directory accessible via the `PATH` environment variable.
 
+#### With Nix flakes
+##### Running nitrocli
+
+Repository comes with a `flake.nix` file, so it can be run directly:
+
+``` bash
```
d-e-s-o, line 120: Seems `sh` is sufficient here, right?

flake.nix
```diff
@@ -16,6 +16,17 @@
...
+          postInstall = ''
+            # copy the manpages
+            install -D -m 0644 ${./.}/doc/nitrocli.1 $out/share/man/man1
+            # make completions
+            mkdir -p $out/share/bash-completion/completions/
+            cargo run --bin=shell-complete bash > $out/share/bash-completion/completions/nitrocli
+            mkdir -p $out/share/zsh/site-functions/
+            cargo run --bin=shell-complete zsh > $out/share/zsh/site-functions/_nitrocli
+            mkdir -p $out/share/fish/vendor_completions.d/
+            cargo run --bin=shell-complete fish > $out/share/fish/vendor_completions.d/nitrocli.fish
```
d-e-s-o, line 28: Would you mind using long options when available? Makes it easier to understand the code.

d-e-s-o (commented):
> Sure. Would you prefer that to be in the same commit or a separate one?

Would you mind just creating a single commit, please? Commit #2 should almost definitely be in the initial commit of the flake, as otherwise it won't work to begin with (right?). We strive to have everything working at each commit. And I don't see a good reason to not contain the man page installation there as well.

I think the larger question is, though, should we actually check in these files here? For Gentoo, for example, they are maintained [out of tree](https://github.com/gentoo/gentoo/tree/master/app-crypt/nitrocli). I understand Ubuntu & Debian are conceptually similar (although all three support overlays of sorts, but getting accepted upstream significantly lowers barriers for users).
It's arguably a bit annoying (because there is no central place to find the recipes for all distros/systems/OSs), but at the same time we don't (tempted to say "can't, within reason") test them properly in the repository anyway, to the point that we have no control over certain aspects (for example a dependent package being renamed upstream) and, hence, could be prone to breakage over time. (I am also not a fan of dumping them into the repository root, but that's arguably a minor aspect)

Not being familiar with Nix, are flakes always managed in a decentralized manner? I found differing information in this respect. Could we, for example, aim to get the flake included [here](https://github.com/NixOS/nixpkgs) as a potentially easier to discover package source?

Can you please see if we can address the license check annotation failures?

VTimofeenko:
Hi!

>Would you mind just creating a single commit, please? 

Sure!

> Commit https://github.com/d-e-s-o/nitrocli/issues/2 should almost definitely be in the initial commit of the flake, as otherwise it won't work to begin with (right?). 

This commit only fixes development shell environment. Nix flakes allow to package the dependencies that are needed only for development in a separate development shell that the user can activate per project.

> We strive to have everything working at each commit. And I don't see a good reason to not contain the man page installation there as well.

Understood

> I think the larger question is, though, should we actually check in these files here? For Gentoo, for example, they are maintained [out of tree](https://github.com/gentoo/gentoo/tree/master/app-crypt/nitrocli). I understand Ubuntu & Debian are conceptually similar (although all three support overlays of sorts, but getting accepted upstream significantly lowers barriers for users).
It's arguably a bit annoying (because t[here](https://github.com/NixOS/nixpkgs) is no central place to find the recipes for all distros/systems/OSs), but at the same time we don't (tempted to say "can't, within reason") test them properly in the repository anyway, to the point that we have no control over certain aspects (for example a dependent package being renamed upstream) and, hence, could be prone to breakage over time. (I am also not a fan of dumping them into the repository root, but that's arguably a minor aspect)
> Not being familiar with Nix, are flakes always managed in a decentralized manner? I found differing information in this respect. Could we, for example, aim to get the flake included here as a potentially easier to discover package source?

Flakes are a bit special. TL;DR - flakes are explicitly a mechanism to embed the recipe to run a program into that program's repository

This is done so that a user of nix-the-package-manager can run on their system without installing that program system wide.

For example, without having neovim or any dependencies installed, I can run:

```
nix run "github:neovim/neovim?dir=contrib/"
```

in my terminal. Nix will resolve all dependencies and start neovim with all the dependencies and parameters specified in [flake.nix](https://github.com/neovim/neovim/blob/master/contrib/flake.nix). This will happen without altering the rest of my system.

> (I am also not a fan of dumping them into the repository root, but that's arguably a minor aspect)
I can move the files to a directory (similar to neovim) and adjust the README - just let me know. 

> Can you please see if we can address the license check annotation failures?

For `flake.nix` - definitely. `flake.lock` has to be json which does not allow for comments. Should I create a `flake.lock.json.license file` - same as `cargo.lock`?

d-e-s-o:
> > I think the larger question is, though, should we actually check in these files here? For Gentoo, for example, they are maintained [out of tree](https://github.com/gentoo/gentoo/tree/master/app-crypt/nitrocli). I understand Ubuntu & Debian are conceptually similar (although all three support overlays of sorts, but getting accepted upstream significantly lowers barriers for users).
> > It's arguably a bit annoying (because t[here](https://github.com/NixOS/nixpkgs) is no central place to find the recipes for all distros/systems/OSs), but at the same time we don't (tempted to say "can't, within reason") test them properly in the repository anyway, to the point that we have no control over certain aspects (for example a dependent package being renamed upstream) and, hence, could be prone to breakage over time. (I am also not a fan of dumping them into the repository root, but that's arguably a minor aspect)
> > Not being familiar with Nix, are flakes always managed in a decentralized manner? I found differing information in this respect. Could we, for example, aim to get the flake included here as a potentially easier to discover package source?
> 
> Flakes are a bit special. TL;DR - flakes are explicitly a mechanism to embed the recipe to run a program into that program's repository
> 
> This is done so that a user of nix-the-package-manager can run on their system without installing that program system wide.
> 
> For example, without having neovim or any dependencies installed, I can run:
> 
> ```
> nix run "github:neovim/neovim?dir=contrib/"
> ```
> 
> in my terminal. Nix will resolve all dependencies and start neovim with all the dependencies and parameters specified in [flake.nix](https://github.com/neovim/neovim/blob/master/contrib/flake.nix). This will happen without altering the rest of my system.

Thanks for the explanation; makes sense to me. For my understanding, though: What if went down the road of getting the flake into [`nixpkgs`](https://github.com/NixOS/nixpkgs) instead? The aforementioned command wouldn't get more complicated, I'd presume, but discoverability may be improved (if that's the "default" repository for people to look for flakes) and "they" may be better eqipped to maintain the recipe over time (for some definition of better).

I am not fundamentally opposed to the inclusion here, just want to make sure we have considered alternatives and ruled them out based on founded arguments.

> > (I am also not a fan of dumping them into the repository root, but that's arguably a minor aspect)
> > I can move the files to a directory (similar to neovim) and adjust the README - just let me know.

Perhaps we could move them into `pkg/nix`?

> > Can you please see if we can address the license check annotation failures?
> 
> For `flake.nix` - definitely. `flake.lock` has to be json which does not allow for comments. Should I create a `flake.lock.json.license file` - same as `cargo.lock`?

Yes, that sounds good to me. Probably ``flake.lock.license`, though.

VTimofeenko:
>Thanks for the explanation; makes sense to me. For my understanding, though: What if went down the road of getting the flake into [nixpkgs](https://github.com/NixOS/nixpkgs) instead? The aforementioned command wouldn't get more complicated, I'd presume, but discoverability may be improved (if that's the "default" repository for people to look for flakes) and "they" may be better eqipped to maintain the recipe over time (for some definition of better).

Nix ecosystem is notorious for overusing the term "nix" to mean several different things and generally being a bit confusing :)

To provide a parallel with python - nixpkgs is like pypi, and a flake is like `pyproject.toml` -  specific to a project or an environment and can contain the pinned dependencies, include development-only dependencies and tools configuration.

I am not that familiar with Rust, but if I understand correctly, pypi = crates.io and `pyproject.toml` = `Cargo.toml`.

In that sense - "having a flake in a project" is not conflicting with "project is available from nixpkgs". Flake approach is a bit more flexible, since the flake can be used in an environment that (for whatever reason) is pinned to an older version of nixpkgs without that project.

d-e-s-o:
Thanks for the explanation.

> In that sense - "having a flake in a project" is not conflicting with "project is available from nixpkgs".

Not conflicting, but it's a different set of trade-offs. The thing is, I don't really want to maintain a nix flake. I am not using it and can't do anything with it, to the degree that I cannot guarantee that it continues working the moment anything changes. And that latter aspect is fundamentally in conflict with what we intend to provide here in this repository. What should I do if a bug report comes in that the flake is no longer working? All I can do is throw my hands in the air and pray that somebody using Nix comes around and fixes the flake.

So from my perspective there are three ways to work around that:
1) Rely on `nixpkgs`. You would land the flake there and that is it (from my perspective). There is no association with this repository and so there is no assumption of folks here having ownership responsibilities. Maintenance falls on you or the next person requiring an updated version and submitting one. On top of that, discoverability may be improved, because (presumably) people can list the contents of `nixpkgs` and become aware of the program to begin with (whereas right now you basically have to know about it to reference it).
2) We have a CI recipe testing that builds on Nix. I don't know how involved that is. I'd think not too much, but haven't done research. We should look for precedent.
3) We strictly version any flakes provided within the repository and treat them as read-only once checked-in (barring potential bug fixes). E.g., you'd provide `nitrocli-0.4.1.nix` that builds `nitrocli` version `0.4.1`. Once `0.4.2` (or whatever) rolls around somebody would need to create an updated flake. `nitrocli-0.4.1.nix` may or may not stay around for a bit longer. In that sense it's similar to 1) (at least how I'd envision it; it's possible you'd have a flake that builds from `master` in `nixpkgs`, but I am not convinced that is wise). That is pretty much by far my least preferred variant, among other things because it (mis)uses this repository (which basically provides snapshots of development over time) as a datastore of sorts.

So after having thought about everything some more, I think that if we want to push through with the change as it is proposed currently, then we'd want CI integration. What are your thoughts?

VTimofeenko:
> The thing is, I don't really want to maintain a nix flake. I am not using it and can't do anything with it, to the degree that I cannot guarantee that it continues working the moment anything changes.

That's perfectly understandable, and I really do not wish to impose this contribution upon you and other maintainers :)

I would like to propose a fourth option - place it in a `contrib/` directory in this repository, and have a README.md there being very explicit about the nature of the flake (and, optionally, other future files in that directory). I believe the arguments for `contrib/` put forward by Drew DeVault in [this blog post](https://drewdevault.com/2020/06/06/Add-a-contrib-directory.html) are very sound.

From technical perspective, a flake is hermetic, it contains a pinned reference to nixpkgs, pinning the rest of the world required for it to run - so unless the nix-the-package-manager's flake interface changes, it will continue to function.

I am considering packaging nitrocli up for nixpkgs, but I am yet to familiarize myself with the way packages are created for nixpkgs - it's quite different from the flakes.

d-e-s-o:
> I would like to propose a fourth option - place it in a `contrib/` directory in this repository, and have a README.md there being very explicit about the nature of the flake (and, optionally, other future files in that directory). I believe the arguments for `contrib/` put forward by Drew DeVault in [this blog post](https://drewdevault.com/2020/06/06/Add-a-contrib-directory.html) are very sound.

On the flip side, to me at least it seems likely that such a directory may encourage sloppy practices, where now folks don't see the need to come up with a proper solution because their hack works for them and it can just be dumped there. Anyway, it's surely a trade-off.

I am not opposed to going down that route and trying it out in this specific case so feel free to adjust the pull request accordingly. That being said, if you integrate with `nixpkgs` down the line please give us a heads up so that we can remove the flake.

VTimofeenko:
I implemented the suggestions from the thread:

- Moved the flake to `contrib/nix`
- Added license information to flake.nix and flake.lock.license
- Adjusted README.md to include new path
- Squashed all commits into one

Sample output:

```
❯ nix run "git+file:///srv/code/forks/nitrocli?dir=contrib/nix" status
Status:
  model:             Nitrokey Storage
...
 ```

(git+file and the path are for my local fork)

Should I add `contrib/README.md` (and a README.md.license) with something along the lines of:

This is the `contrib/` directory of nitorcli project. Files located here:
- Are provided by the community of the project
- Are not official part of the project 
- Are provided without any warranty

?

**Comment by d-e-s-o** [comment]
Thanks!

> Should I add `contrib/README.md` (and a README.md.license) with something along the lines of:

That would be great!

I hope to take a closer look this weekend.

File: contrib/nix/flake.nix [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+          '';
+        };
+
+        apps.default = utils.lib.mkApp {
+          drv = packages.default;
+          name = "nitrocli";
+        };
+
+        devShell = with pkgs; mkShell {
+          buildInputs = [ cargo rustc rustfmt pre-commit rustPackages.clippy hidapi ];
```
d-e-s-o, line 40: For my understanding, why is `rustfmt` needed here?

VTimofeenko: `rustfmt` just came from the [template](https://github.com/NixOS/templates/tree/master/rust).

> Is this file also suitable for installing all dependencies for development?

Correct, `devShell` contents can be triggered manually (`nix develop`) or through `direnv` to pull in development-only dependencies.

d-e-s-o:
Great stuff, thanks! I made some minor changes that hopefully didn't break anything (:o) and merged the patch.

